PR #14, opened 2017-06-08
General UI improvements

anehx:
* Indicate loading state
* Better lazy loading
* Better task selection
* Better absences mask


anehx:
**TODO:** @anehx add missing docstrings before requesting review / merge

app/components/task-selection/component.js
```diff
@@ -27,215 +33,210 @@ export default Component.extend({
...
+      /* istanbul ignore next */
+      tasks
+        .then((data) => {
+          let re = new RegExp(`.*${search}.*`, 'i')
+
+          return asyncResults(data.filter((i) => re.test(i.get('name'))))
+        })
+        .catch(() => {
+          return asyncResults([])
+        })
```
czosel: 1. The three sources (task, project, customer) are practically identical - think you can pull a function out?
2. Why not use async/await here?

app/components/weekly-overview-day/styles.scss
```diff
@@ -1,9 +1,10 @@
 & {
-  height: 100%;
+  height: 0%;
```
czosel: Nitpick: just `0`?

app/index/controller.js
```diff
@@ -209,21 +233,65 @@ export default Controller.extend({
...
       return {
         day: d,
-        active: d.isSame(current, 'day'),
-        worktime: reports.filter((r) => r.get('date').isSame(d, 'day')).reduce((val, cur) => {
-          return val.add(cur.get('duration'))
+        active: d.isSame(date, 'day'),
+        absence: !!absences.length,
+        workday: this.get('workdays').includes(d.isoWeekday()),
+        worktime: [ ...reports, ...absences ].reduce((val, dur) => {
+          return val.add(dur)
```
czosel: sure that this shouldn't be 
```
val.add(dur)
return val
```
If yes, what's the return value of moment.add?

app/login/controller.js
```diff
@@ -0,0 +1,52 @@
...
+   * @method authenticate
+   * @param {String} username The username
+   * @param {String} password The password
+   * @public
+   */
+  authenticate: task(function* (username, password) {
+    try {
+      yield this.get('session').authenticate('authenticator:application', { username, password })
+
+      this.store.unloadAll()
```
czosel: :+1:

app/me/template.hbs
```diff
@@ -26,11 +26,11 @@
                 <td class="text-right">{{model.activeEmployment.percentage}}%</td>
               </tr>
               <tr>
-                <td>Wortime per day:</td>
+                <td>Worktime per day:</td>
```
czosel: finally! :rofl: :sweat_smile:

app/services/tracking.js
```diff
@@ -0,0 +1,198 @@
...
+   *
+   * @property {EmberConcurrency.Task} filterCustomers
+   * @public
+   */
+  filterCustomers: task(function* () {
+    yield timeout(500)
+
+    let customers = yield this.get('store').query('customer', {})
+
+    return customers
```
czosel: Could this be rewritten as `return yield ... `?

app/services/tracking.js
```diff
@@ -0,0 +1,198 @@
...
+    if (!project) {
+      throw new Error('No project selected')
+    }
+
+    yield timeout(500)
+
+    let tasks = yield this.get('store').query('task', { project })
+
+    return tasks
+  }).restartable(),
```
czosel: Again, maybe you can pull a common function out of the three filters?

bower.json
```diff
@@ -3,6 +3,7 @@
   "dependencies": {
     "adcssy": "adfinis-sygroup/adcssy#0.9.1",
     "font-awesome": "^4.7.0",
-    "nouislider": "^8.0.1"
+    "nouislider": "^8.0.1",
+    "corejs-typeahead": "adfinis-forks/typeahead.js#master"
```
czosel: Don't forget to open a PR ;)

anehx: Nope.. No PR because the functionality of the fork is not a common needed thing. It's very custom to our use case.

app/services/tracking.js
anehx: I wouldn't do that.. Just because there are 3 similar functions which filter different models, we don't have to put that into one function. Every filter is custom to the model so why put effort into that?

app/index/controller.js
anehx: Why do you think this is wrong? https://momentjs.com/docs/#/manipulating/add/ => return value is the mutated moment object..

app/components/weekly-overview-day/styles.scss
anehx: No. Height is 0% so it gets animated when the height changes to [x]%

app/components/task-selection/component.js
anehx: 1. Again: Worth the effort? Only the structure is similar. Functionality is completely different and uses different filters which would create a pretty ugly function to cover all cases..
2. This is a ember-concurrency task not a promise. You can't use promise functionality there.

app/services/tracking.js
anehx: Done!

app/index/controller.js
czosel: Ah, nice. I thought so because in `app/index/controller.js#135` you wrote it in a mutating style - or is this because it's a duration?

app/components/weekly-overview-day/styles.scss
czosel: Right!

app/index/controller.js
anehx: Ah well I see, inconsistency! :grin: Duration acts the same, gonna fix it.

anehx: Done!

app/components/task-selection/component.js
czosel: 1. I overlooked that you're calling `perform` without any argument in the first source, which would make a generic version indeed much harder to read. But how about pulling this one out? 
```
const regexFilter = (data, asyncResults) => {
  let re = new RegExp(`.*${search}.*`, 'i')
  return asyncResults(data.filter((i) => re.test(i.get('name'))))
}
```
Might also not be worth it cause you have to pass `asyncResults` - i guess it's up to you, i just dont like too much duplication.

anehx: I already considered this one but decided not to extract it because we would need to pass `asyncResults`. "Duplication" is the lesser evil IMHO so I'll let this be.

czosel: Fair enough!

app/index/controller.js
czosel: :+1: The one-liner is much easier on the eyes :)

PR #17, opened 2017-06-16
Edit activities

anehx:
* Show edit form next to the list instead of a modal
* Allow editing, deleting and adding of activity blocks
* Add precision, max and min value to time- and durationpicker
* Add possibility to pass only a number (milliseconds) into the `formatDuration` function


czosel:
Added some minor nitpicks, otherwise LGTM!

anehx:
I found a similar issue [here](https://github.com/jmcriffey/babel-istanbul/issues/63) so I tried locking babel-core to version 6.7.4 but it did not do anything..

anehx:
@kaldras, @czosel I applied the requested changes. ATM I can't fix the coverage (see why [here](https://github.com/kategengler/ember-cli-code-coverage/issues/124)) so we may need to settle with 99% :cry:

PR #18, opened 2017-06-18
Set document title if an activity is active

anehx:


app/services/tracking.js
```diff
@@ -54,9 +68,99 @@ export default Service.extend({
...
+
+      if (this.get('activity.task.content')) {
+        let c = this.get('activity.task.project.customer.name')
+        let p = this.get('activity.task.project.name')
+        let t = this.get('activity.task.name')
+
+        task = `${c} > ${p} > ${t}`
+      }
+      else {
+        task = 'Unknown Task'
```
kaldras: Why not define task as 'Unknown Task', then you don't need the else part.

anehx: Makes sense..

PR #20, opened 2017-06-29
Feature activity history

kaldras:
When selecting a customer for an activity or a report, request history entries and display them at the top. A history entry fills customer > project > task with one selection.

### DIFF
--- a/app/components/task-selection/component.js
+++ b/app/components/task-selection/component.js
@@ -6,11 +6,12 @@
 import Component  from 'ember-component'
 import computed   from 'ember-computed-decorators'
 import service    from 'ember-service/inject'
+import RSVP       from 'rsvp'
 import hbs        from 'htmlbars-inline-precompile'
 import { typeOf } from 'ember-utils'
 
 const FORMAT = (obj) => typeOf(obj) === 'instance' ? obj.get('name') : ''
-const SUGGESTION_TEMPLATE = hbs`{{#if model.start}}<i class="fa fa-history"></i> {{/if}}{{model.name}}`
+const SUGGESTION_TEMPLATE = hbs`{{#if model.duration}}<i class="fa fa-history"></i> {{/if}}{{model.name}}`
 
 const regexFilter = (data, term, key) => {
   let re = new RegExp(`.*${term}.*`, 'i')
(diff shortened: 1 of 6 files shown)


app/components/task-selection/component.js
```diff
@@ -112,6 +121,18 @@ export default Component.extend({
       return task && task.get('project.customer') || this.get('_customer')
     },
     set(value) {
+      /**
+       * It is also possible a activity/report was selected from the history.
+       * An activity/report has a duration property, check for this one.
+       */
+      if (value !== null && value.get('duration')) {
```
czosel: Why not just `if (value && value.get('duration')) {`

app/components/task-selection/component.js
```diff
@@ -112,6 +121,18 @@ export default Component.extend({
...
+        this.set('task', value.get('task'))
+        this.set('_customer', value.get('task.project.customer'))
+        this.set('_project', value.get('task.project'))
+
+        return this.get('_customer')
```
czosel: Why do you have to return the customer here?

app/components/task-selection/component.js
kaldras: True that and changed

app/components/task-selection/component.js
kaldras: According to the Ember computed documentation, the computed setter should return the new value. As this is the customer property, we return the customer.

czosel: :+1: didn't know

app/models/activity.js
```diff
@@ -99,5 +99,23 @@ export default Model.extend({
...
+   * @type {String}
+   * @public
+   */
+  @computed('task')
+  name(task) {
+    let taskName = task.get('name')
+    let projectName = task.get('project.name')
+    let customerName = task.get('project.customer.name')
+
+    return `${customerName} > ${projectName} > ${taskName}`
```
anehx: What happens if the task isn't set? This is theoretically not possible but I'd still handle that case somehow..

app/models/report.js
```diff
@@ -81,5 +82,23 @@ export default Model.extend({
...
+   * @public
+   */
+  @computed('task')
+  name(task) {
+    let taskName = task.get('name')
+    let projectName = task.get('project.name')
+    let customerName = task.get('project.customer.name')
+
+    return `${customerName} > ${projectName} > ${taskName}`
+  }
```
anehx: Same as above

app/models/activity.js
kaldras: Yes, will be handled now.

app/models/report.js
kaldras: Also done

kaldras:
Was forked and finished by #30 - closing

PR #22, opened 2017-07-19
Filter reports by user

czosel:
Depends on #25.


sliverc:
Looks good. This could actually already be merged as the old api would simply ignore the user field.

Merge conflicts need to be solved first though.

czosel:
True! Fixed the merge conflicts.

PR #23, opened 2017-07-19
Rename "reports" to "timesheet"

czosel:


andreabettich:
Don't know why the coverage decreased  but it looks good.

PR #36, opened 2017-08-07
Custom report export

andreabettich:
Give the possibility to configure the analysis export endpoints.
This PR is based on the #35 PR/Code, so please review #35 first.

### DIFF
--- a/app/analysis/controller.js
+++ b/app/analysis/controller.js
@@ -1,5 +1,6 @@
 import Ember from 'ember'
 import Controller from 'ember-controller'
+import config from '../config/environment'
 
 const { testing } = Ember
 
@@ -22,7 +23,7 @@ export default Controller.extend({
   page: 1,
 
   sort: '-date',
-
+  exportLinks: config.APP.REPORTEXPORTS,
   actions: {
     download(type, filters) {
       let target = `${downloadUrl(filters)}&file_type=${type}`
--- a/app/analysis/controller.js
+++ b/app/analysis/controller.js
@@ -4,17 +4,12 @@ import config from '../config/environment'
 
 const { testing } = Ember
 
-const downloadUrl = filters => {
-  /* eslint-disable camelcase */
-  let base = '/api/v1/reports/export?'
-  return (
-    base +
-    Object.keys(filters)
-      .map(key => {
-        return `${key}=${filters[key]}`
-      })
-      .join('&')
-  )
+const downloadUrl = (url, filters) => {
+  return `${url}&${Object.keys(filters)
+    .map(key => {
+      return `${key}=${filters[key]}`
+    })
+    .join('&')}`
 }
 
 export default Controller.extend({
@@ -25,8 +20,8 @@ export default Controller.extend({
   sort: '-date',
   exportLinks: config.APP.REPORTEXPORTS,
   actions: {
-    download(type, filters) {
-      let target = `${downloadUrl(filters)}&file_type=${type}`
+    download(url, filters) {
+      let target = downloadUrl(url, filters)
 
       if (testing) {
         return
--- a/app/analysis/controller.js
+++ b/app/analysis/controller.js
@@ -4,7 +4,7 @@ import config from '../config/environment'
 
 const { testing } = Ember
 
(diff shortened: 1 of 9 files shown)


app/analysis/controller.js
```diff
@@ -1,19 +1,15 @@
...
-  return (
-    base +
-    Object.keys(filters)
-      .map(key => {
-        return `${key}=${filters[key]}`
-      })
-      .join('&')
-  )
+const downloadUrl = (url, filters) => {
+  return `${url}&${Object.keys(filters)
```
czosel: Looks mad, but is fine :smile:

app/analysis/controller.js
```diff
@@ -22,10 +18,10 @@ export default Controller.extend({
   page: 1,
 
   sort: '-date',
-
+  exportLinks: config.APP.REPORTEXPORTS,
   actions: {
-    download(type, filters) {
-      let target = `${downloadUrl(filters)}&file_type=${type}`
+    download(url, filters) {
+      let target = downloadUrl(url, filters)
```
czosel: What do you think about renaming the method to something like `appendFilters`?

andreabettich: Yes, would be more informative.

andreabettich: Changed it on #a582355

config/environment.js
```diff
@@ -82,5 +82,21 @@ module.exports = function(environment) {
...
+  /* global process*/
+  try {
+    let additionalReportExports = JSON.parse(process.env.timed_report_export)
+    if (additionalReportExports) {
+      ENV.APP.REPORTEXPORTS = ENV.APP.REPORTEXPORTS.concat(
+        additionalReportExports
+      )
+    }
+  } catch (e) {
+    /* eslint no-console: ["error", { allow: ["error"] }] */
```
czosel: There are easier ways to write this, for example:
`// eslint-disable-next-line no-console`

See http://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments

config/environment.js
```diff
@@ -82,5 +82,21 @@ module.exports = function(environment) {
 
   }
 
+  /* global process*/
+  try {
+    let additionalReportExports = JSON.parse(process.env.timed_report_export)
+    if (additionalReportExports) {
+      ENV.APP.REPORTEXPORTS = ENV.APP.REPORTEXPORTS.concat(
```
czosel: `ENV.APP.REPORTEXPORTS = [ ...ENV.APP.REPORTEXPORTS, ...additionalReportExports ]`

config/environment.js
```diff
@@ -82,5 +82,21 @@ module.exports = function(environment) {
 
   }
 
+  /* global process*/
+  try {
+    let additionalReportExports = JSON.parse(process.env.timed_report_export)
+    if (additionalReportExports) {
```
czosel: You might also want to check if it's an array.

czosel:
I just discussed this a little further with @sliverc, we thought that the following should still be added:

When you develop locally, it should be easy to set up a custom environment. For that, we'd like to support setting the environment in an `.env` file. That file will not be checked in version control, but a command to set it up (`echo .. > .env`) should be added to the README in our private repo. In the README of this repo there should be some documentation on what this feature is doing and how to use it.

config/environment.js
andreabettich: Damn, that looks 1000x sexier :smile:

config/environment.js
andreabettich: Changed it

config/environment.js
```diff
@@ -70,5 +82,24 @@ module.exports = function(environment) {
 
   }
 
+  /* global process*/
+  try {
+    let envReportExports = process.env.TIMED_REPORT_EXPORT
+    if (envReportExports) {
+      let additionalReportExports = JSON.parse(envReportExports)
+      if (additionalReportExports && additionalReportExports instanceof Array) {
```
czosel: I think `Array.isArray()` is better practice than `instanceof`

andreabettich: Didn't even know that existed. Nice!

PR #39, opened 2017-08-08
Datepicker clear symbol

andreabettich:


app/components/sy-datepicker/template.hbs
```diff
@@ -11,6 +11,9 @@
       value={{moment-format value 'DD.MM.YYYY' allow-empty=true}}
       placeholder={{placeholder}}
     >
+    <span class="clear" {{action 'clear'}}>
+      <i class="fa fa-times" aria-hidden="true"></i>
```
anehx: Why not use `{{fa-icon 'times' click=(action 'clear')}}` here?

andreabettich: was not aware that we had such a addon included

app/components/sy-datepicker/template.hbs
```diff
@@ -11,8 +11,8 @@
       value={{moment-format value 'DD.MM.YYYY' allow-empty=true}}
       placeholder={{placeholder}}
     >
-    <span class="clear" {{action 'clear'}}>
-      <i class="fa fa-times" aria-hidden="true"></i>
+    <span class="clear">
+      {{fa-icon 'times' click=(action 'clear')}}
     </span>
```
anehx: Why do you still need the `span`?

PR #46, opened 2017-08-15
Overlaping activities

anehx:
* User gets a warning when trying to generate a timesheet with an activity which overlaps the day
* Overlapping activities and blocks are highlighted
* Help texts for the user
* Field `start-datetime` was renamed in the backend to `date` for better understanding

### DIFF
--- a/app/index/activities/route.js
+++ b/app/index/activities/route.js
@@ -101,9 +101,22 @@ export default Route.extend({
      * @public
      */
     generateReportsCheck() {
-      this.get('controller.activities').filterBy('task.id', undefined).length
-        ? this.set('controller.showWarning', true)
-        : this.send('generateReports')
+      let hasUnknown = this.get('controller.activities').find(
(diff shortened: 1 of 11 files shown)


app/index/activities/route.js
```diff
@@ -101,9 +101,20 @@ export default Route.extend({
...
-      this.get('controller.activities').filterBy('task.id', undefined).length
-        ? this.set('controller.showWarning', true)
-        : this.send('generateReports')
+      let hasUnknown = this.get('controller.activities').find(a => {
+        return !a.get('task.id')
+      })
+
+      let hasOverlapping = this.get('controller.activities').find(a => {
+        return a.get('overlaps')
+      })
```
czosel: How about `!!this.get('controller.activities').findBy('overlaps')`
and 
`!!this.get('controller.activities').findBy('task.id', false)`

PR #50, opened 2017-08-15
sort users in user selection

andreabettich:


czosel:
You might be able to implement this as a one-liner with the help of https://github.com/DockYard/ember-composable-helpers#sort-by

andreabettich:
@czosel Thanks, looks cleaner with the helper you linked. I'll look into it and change it

anehx:
Fixed in #60

PR #53, opened 2017-08-17
Improved controlling functions

anehx:
* URL filters for reschedule and analysis
* More filters
* Better UI for rescheduling
* New function to verify reports (rescheduling)


app/analysis/controller.js
```diff
@@ -41,16 +38,25 @@ const AnalysisController = Controller.extend({
...
-  getTarget(url, filters) {
+  getTarget(url) {
+    let filters = this.get('queryParams').reduce((obj, key) => {
+      let val = this.get(key)
+
+      if (val !== null && val !== undefined) {
+        obj[key] = val
+      }
+
+      return obj
```
czosel: With object spread you should be able to simplify this:
```
let val = this.get(key)
return val ? {...obj, [key]: val} : obj
```

app/components/report-filter/component.js
```diff
@@ -0,0 +1,157 @@
...
+   * @method _setDefaultFilters
+   * @param {Object} initial Initial values
+   * @private
+   */
+  _setDefaultFilters(initial = {}) {
+    let filters = EmberObject.create(
+      this.get('_enabledFilters').reduce((obj, key) => {
+        obj[key] = initial[key] || FILTERS[key].default
+
+        return obj
```
czosel: Same here (object spread)

app/components/report-filter/template.hbs
```diff
@@ -0,0 +1,102 @@
+<div class="grid grid--2of12">
+  {{#task-selection
```
czosel: Don't we also need to check here if `customer`, `project` or `task` are enabled?

app/mixins/report-filter-controller.js
```diff
@@ -0,0 +1,154 @@
...
+     *
+     * @method applyFilter
+     * @param {Object} filters The filters to apply
+     * @public
+     */
+    applyFilter(filters) {
+      let parsed = Object.keys(filters).reduce((obj, key) => {
+        obj[underscore(key)] = filters[key]
+
+        return obj
```
czosel: Same

app/reschedule/route.js
```diff
@@ -1,3 +1,129 @@
...
+    /**
+     * Save a report row
+     *
+     * @method saveReport
+     * @param {Report} report The report to save
+     * @public
+     */
+    async saveReport(report) {
+      try {
+        this.send('loading')
```
czosel: OT: Do you see a way how we could get rid of the annoying `loading` and `finished` boilerplate? Does ember-concurrency offer anything for this case?

app/reschedule/route.js
```diff
@@ -1,3 +1,129 @@
...
+        let queryString = Object.keys(params)
+          .reduce((arr, key) => {
+            let val = params[key]
+
+            if (val !== null && val !== undefined) {
+              arr.push(`${key}=${val}`)
+            }
+
+            return arr
+          }, [])
```
czosel: You're cleaning up `params` and mapping it to a query string here. I prefer to compose two functions in cases like this, cause its more readable and reusable:
```
let queryString = toQueryString(clean(params))

const clean = obj => {
  return Object.keys(obj)
    .filter(key => Boolean(obj[key]))
    .reduce((out, key) => ({ ...out, [key]: obj[key] }), {})
}
const toQueryString = obj => Object.keys(obj).map(key => `${key}=${obj[key]}`).join('&')
```
The two functions might be best placed in some utility class - do we have a place for this?

app/reschedule/template.hbs
```diff
@@ -1,54 +1,66 @@
...
-        <th>Date</th>
-        <th>Dur.</th>
-        <th>Customer / Project / Task</th>
-        <th>Flags</th>
-        <th>User</th>
-        <th>Comment</th>
-        <th>Action</th>
+        <td colspan="11" class="text-center">
+          {{#if (is-array model)}}
+            I didn't find any results - try broadening your filters!
```
czosel: Not sure if "broaden filters" is something an english-speaking person would say. I've heard of "broaden search", and i could imagine that "loosen filters" or "changing filters" would also be ok.

app/reschedule/route.js
anehx: Not at the moment..

app/analysis/controller.js
```diff
@@ -3,66 +3,41 @@ import Controller from 'ember-controller'
...
 
   /**
    * The current JWT token
    *
-   * @property {string} token
+   * @property {string} jwt
    * @public
    */
-  @oneWay('session.data.authenticated.token') token: null,
+  @oneWay('session.data.authenticated.token') jtw: null,
```
czosel: `jwt`?

anehx: Whoops..

app/analysis/controller.js
```diff
@@ -3,66 +3,41 @@ import Controller from 'ember-controller'
...
-
-      if (val !== null && val !== undefined) {
-        obj[key] = val
-      }
-
-      return obj
-    }, {})
+    let queryString = toQueryString(
+      cleanParams(this.getProperties(...this.get('queryParams'), 'jwt'))
+    )
```
czosel: :+1:

tests/unit/utils/url-test.js
```diff
@@ -0,0 +1,36 @@
...
+      5: 'test'
+    }
+
+    let result = cleanParams(params)
+
+    expect(result).to.deep.equal({
+      1: '',
+      4: 0,
+      5: 'test'
+    })
```
czosel: Unit tests for simple functions are fun :smile:

PR #57, opened 2017-08-18
Ignore archived customers, projects and task on the tracking page

anehx:
#53 needs to be merged first, so I can apply the archived property for the refactored pages..


app/components/task-selection/component.js
```diff
@@ -235,13 +243,15 @@ export default Component.extend({
...
+      let customers =
+        fnCustomer.get('last') || fnCustomer.perform(this.get('archived'))
 
       let requests = { customers }
 
       if (history) {
         let fnHistory = this.get('tracking.recentTasks')
-        requests.history = fnHistory.get('last') || fnHistory.perform()
+        requests.history =
+          fnHistory.get('last') || fnHistory.perform(this.get('archived'))
```
czosel: It might make sense to define something like
```
const performCached = (fn, ...args) => fn.get('last') || fn.perform(...args)
```

PR #66, opened 2017-08-22
Allow user to select between page sizes 10, 20, 50 and 100

anehx:


app/components/pagination-pager/template.hbs
```diff
@@ -0,0 +1,52 @@
...
+      <a href="{{nextUrl}}" {{action 'next'}}>{{{pagerNext}}}</a>
+    </li>
+
+    {{#if pagerLast}}
+      <li class="last {{if isLastDisabled 'disabled'}}">
+        <a href="{{lastUrl}}" {{action 'last'}}>{{{pagerLast}}}</a>
+      </li>
+    {{/if}}
+  {{/unless}}
+</ul>
```
czosel: The first part is exactly [this template](https://github.com/knownasilya/pagination-pager/blob/master/app/templates/components/pagination-pager.hbs), correct? Isn't there any way of including it instead of repeating it? If not, i'd suggest commenting that where this comes from and why it was extended.

PR #75, opened 2017-09-01
Show the absence type or the holiday name in the weekly overview on hover

anehx:


app/index/controller.js
```diff
@@ -15,6 +15,12 @@ import { scheduleOnce } from 'ember-runloop'
 
 const { testing } = Ember
 
+const MODEL_NAME_MAPPING = {
+  report: 'reports',
+  absence: 'absences',
+  'public-holiday': 'holidays'
```
sliverc: It is not obvious here why public holiday is a hard coded string here and report and absence are not.

czosel: not obvious for someone who doesn't know JS :-P

czosel: (object property keys have to be valid identifiers in JS. Something containing dashes isn't: https://stackoverflow.com/questions/1661197/what-characters-are-valid-for-javascript-variable-names)

PR #81, opened 2017-09-01
Fixed boolean filters

anehx:


app/components/report-filter/component.js
```diff
@@ -9,8 +9,9 @@ import computed from 'ember-computed-decorators'
 
 const DATE_FORMAT = 'YYYY-MM-DD'
 
+// Options can be '', '1' or '2' from a select box
```
sliverc: shouldn't this be '', '0' or '1'?

PR #83, opened 2017-09-04
Some minor improvements

anehx:
* Better timesheet / attendances styling for small devices
* Better validation for duration and timepicker
* Better highlighting of the selected day in the weekly overview

### DIFF
--- a/app/components/sy-timepicker/component.js
+++ b/app/components/sy-timepicker/component.js
@@ -23,12 +23,14 @@ export default Component.extend({
    * @public
    */
   init() {
+    let value = this.get('value') || moment()
+
     if (!this.get('min')) {
-      this.set('min', moment(this.get('value')).set({ h: 0, m: 0 }))
+      this.set('min', moment(value).set({ h: 0, m: 0 }))
     }
 
     if (!this.get('max')) {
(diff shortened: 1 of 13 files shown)


app/components/sy-timepicker/component.js
```diff
@@ -70,7 +72,15 @@ export default Component.extend({
    * @private
    */
   _add(h, m) {
-    return moment(this.get('value')).add({ h, m })
+    let base = this.get('value')
+
+    if (!base) {
+      base = [h, m].any(n => n < 0)
+        ? this.get('max').add(1, 'minute')
+        : this.get('min')
```
czosel: I'm not sure if i get this. Why is the base`max + 1min` if `h` or `m` is negative?

czosel: Oh that's to make sure you're not entering "yesterday" by clicking the down arrow, right?

anehx: If we don't have a current value and press arrow down, which either subtracts 15 minutes or one hour, we subtract it from the current max which is normally 23:59 if not given. So we need to add a minute to get a valid value.

czosel: :+1:

PR #84, opened 2017-09-04
Added sort headers for rescheduling and styled the analysis table better

anehx:


app/analysis/template.hbs
```diff
@@ -46,6 +50,27 @@
...
+        <td>
+          {{#if report.notBillable}}
+            {{fa-icon 'check-square-o'}}
+          {{else}}
+            {{fa-icon 'square-o'}}
+          {{/if}}
+        </td>
+        <td>
+          {{#if report.verifiedBy}}
+            {{fa-icon 'check-square-o' title=(concat 'Verified by ' report.verifiedBy.username)}}
```
sliverc: I think it would be better to simply show the verifier instead of a check box. The check box was mainly for usage in reschedule to set current user but in analysis we want to know who was it.

app/analysis/template.hbs
```diff
@@ -46,6 +50,21 @@
         <td>{{report.task.project.name}}</td>
         <td>{{report.task.name}}</td>
         <td>{{report.comment}}</td>
+        <td>
+          {{#if report.review}}
+            {{fa-icon 'check-square-o'}}
+          {{else}}
+            {{fa-icon 'square-o'}}
+          {{/if}}
```
czosel: How about a new little component?
```
{{sy-checkmark checked=report.review}}
```

app/analysis/template.hbs
```diff
@@ -33,6 +34,9 @@
       {{#sort-header update=(action (mut ordering)) current=ordering by='task__project__name'}}Project{{/sort-header}}
       {{#sort-header update=(action (mut ordering)) current=ordering by='task__name'}}Task{{/sort-header}}
       {{#sort-header update=(action (mut ordering)) current=ordering by='comment'}}Comment{{/sort-header}}
+      <th>Needs review</th>
+      <th>Not billable</th>
+      <th>Verifier</th>
```
czosel: Doesn't the backend support sorting those columns?

anehx: No. It does not really make sense to sort by booleans here so we decided not to let them be sortable.

app/analysis/template.hbs
anehx: Done!

app/analysis/template.hbs
anehx: Done!

app/analysis/template.hbs
czosel: Alright, thanks!

anehx:
@czosel Might be good if we defined a standard here instead of just everyone using some different "standard"..

czosel:
Definitely, lets bring this up during the next team meeting.

app/analysis/template.hbs
sliverc: As we changed Verified to Verifier it actually would make sense to be able to sort by verified_by. On second thought label might be better as Verified by.
I adjust the backend to allow this.

sliverc:
LGTM now. Thanks.

PR #89, opened 2017-09-04
Fixed filtering by user on the tracking pages

anehx:


app/index/reports/controller.js
```diff
@@ -38,10 +38,14 @@ export default Controller.extend({
...
    * @public
    */
-  @computed('_allReports.@each.{date,isNew,isDeleted}', 'model')
-  reports(reports, day) {
+  @computed('_allReports.@each.{user,date,isNew,isDeleted}', 'model', 'user')
+  reports(reports, day, user) {
     let reportsToday = reports.filter(r => {
-      return r.get('date').isSame(day, 'day') && !r.get('isDeleted')
+      return (
+        (!r.get('user.id') || r.get('user.id') === user.get('id')) &&
```
czosel: This is the only place where you also check for `r.get('user.id')` - why?

app/index/controller.js
```diff
@@ -87,15 +87,19 @@ export default Controller.extend({
...
-  @computed('date', '_allActivities.@each.{start,isDeleted}')
-  _activities(day, activities) {
+  @computed('date', '_allActivities.@each.{start,user,isDeleted}', 'user')
+  _activities(day, activities, user) {
     let activitiesThen = activities.filter(a => {
-      return a.get('start').isSame(day, 'day') && !a.get('isDeleted')
+      return (
+        a.get('start').isSame(day, 'day') &&
+        a.get('user.id') === user.get('id') &&
+        !a.get('isDeleted')
```
czosel: For my taste, the filter is duplicated so much that it might be forth factoring it out. For lack of a better name, how about
`activities.filter(currentPersonal)`

app/index/reports/controller.js
anehx: Because this is the only case in which we can have unsaved items, without a user.

czosel: Ah, i think i also know what part of the UI this belongs to now. Thanks!

app/index/controller.js
anehx: I personally don't like extracting a filter which filters different entities..

PR #93, opened 2017-09-07
Show the progress of a project and task

anehx:
This PR implements a small tooltip which shows the estimated and spent time in the comboboxes of projects and tasks.


app/services/metadata-fetcher.js
```diff
@@ -0,0 +1,65 @@
...
+   * @method fetchSingleRecordMetadata
+   * @param {String} type The type to fetch
+   * @param {Number} id The id of the object to fetch
+   * @return {Object} An object with the parsed metadata
+   * @public
+   */
+  fetchSingleRecordMetadata: task(function*(type, id) {
+    /* istanbul ignore next */
+    if (!id) {
+      throw new Error('Project ID is missing')
```
kaldras: Why `Project ID is...`? Probably use the `type` here?

anehx: Leftover from when it was only supported for projects.. Thanks!

PR #95, opened 2017-09-08
Fix activity editing and remove some unwanted sideeffects

anehx:
* Add explicit types for all buttons
* Proper validations and changesets for activity blocks
* Some minor fixes for the time-/durationpicker

### DIFF
--- a/app/analysis/template.hbs
+++ b/app/analysis/template.hbs
@@ -77,6 +77,7 @@
   <p>
     {{#each exportLinks as |link index|}}
       <button {{action 'download' link.url}}
+        type="button"
         disabled={{gt model.length 4000}}
         class="btn btn-default"
         data-test-download-file={{index}}>{{link.label}}</button>
--- a/app/components/sy-timepicker/component.js
+++ b/app/components/sy-timepicker/component.js
@@ -7,6 +7,7 @@ import Component from 'ember-component'
 import computed from 'ember-computed-decorators'
 import moment from 'moment'
 import { padStart } from 'ember-pad/utils/pad'
+import { debounce } from 'ember-runloop'
 
 /**
  * Timepicker component
@@ -103,7 +104,11 @@ export default Component.extend({
    * @private
    */
   _addMinutes(value) {
-    this._change(this._add(0, value))
+    let newValue = this._add(0, value)
+
+    if (this._validate(newValue)) {
+      this._change(newValue)
+    }
   },
 
   /**
@@ -114,7 +119,11 @@ export default Component.extend({
    * @private
    */
   _addHours(value) {
-    this._change(this._add(value, 0))
+    let newValue = this._add(value, 0)
+
+    if (this._validate(newValue)) {
+      this._change(newValue)
+    }
   },
 
   /**
@@ -125,9 +134,7 @@ export default Component.extend({
    * @private
    */
   _change(value) {
-    if (value === null || this._validate(value)) {
-      this.get('attrs.on-change')(value)
-    }
+    this.get('attrs.on-change')(value)
   },
 
   /**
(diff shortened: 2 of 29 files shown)


app/validators/moment.js
```diff
@@ -0,0 +1,38 @@
...
+      if (newValue <= gtVal) {
+        valid = false
+      }
+    }
+    if (options.lt) {
+      let ltVal = content.get(options.lt) || changes[options.lt]
+
+      if (newValue >= ltVal) {
+        valid = false
+        valid = false
```
czosel: copy pasta

app/analysis/template.hbs
```diff
@@ -77,6 +77,7 @@
   <p>
     {{#each exportLinks as |link index|}}
       <button {{action 'download' link.url}}
+        type="button"
```
czosel: Out of curiosity: Why specify the type in this case?

anehx: Because I don't want any side effects like e.g pressing enter in the filter form and triggering a download with it.. IMHO submit buttons should only be allowed in forms and only one per form.

app/components/sy-timepicker/component.js
```diff
@@ -214,11 +229,12 @@ export default Component.extend({
...
-          this._change(null)
-        }
-
-        this._change(this._set(h, m))
+        debounce(
+          this,
+          this._change,
+          [h, m].some(isNaN) ? null : this._set(h, m),
+          this.get('delay')
+        )
```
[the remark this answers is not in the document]
anehx: If I have a value like 10:30 and go want to change it to 11:30 I'd try to delete the zero of the 10 and enter a one. If you did this without a debounce, the value would change to 01:30 as soon as you delete the zero.

app/components/sy-timepicker/component.js
```diff
@@ -103,7 +104,11 @@ export default Component.extend({
    * @private
    */
   _addMinutes(value) {
-    this._change(this._add(0, value))
+    let newValue = this._add(0, value)
+
+    if (this._validate(newValue)) {
```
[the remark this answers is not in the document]
anehx: I'd call it `_checkValidity` since `isValid` sounds like a property to me which it clearly isn't. I'll change this if you agree?

app/analysis/template.hbs
czosel: Damn, i didn't know the default type is `submit`. Makes total sense.

anehx: Yep that took me by surprise too.. I had a form which deleted entries when pressing enter.

app/components/sy-timepicker/component.js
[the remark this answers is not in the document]
czosel: And with the debounce that happens after 500ms, correct? Isn't it better UX to do formatting/validation when the input looses focus?

app/components/sy-timepicker/component.js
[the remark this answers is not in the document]
czosel: `checkValidity` also sounds like the function is "doing" something complicated. Since your function is pure, it behaves like a computed property, so i'd be fine with calling it `isValid`.
I understand your point too though - just leave it as it is if you don't like `isValid` ;)

app/components/sy-timepicker/component.js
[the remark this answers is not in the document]
anehx: Seems fair. Gonna fix it!

app/components/sy-timepicker/component.js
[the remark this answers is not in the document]
anehx: Valid point :wink: Done!

app/index/activities/edit/route.js
```diff
@@ -103,12 +139,26 @@ export default Route.extend(RouteAutostartTourMixin, {
...
+          this.get('controller.blocks')
+            .filter(({ changeset, model }) => {
+              return changeset.get('isDirty') || model.get('isDeleted')
+            })
+            .map(async ({ changeset, model }) => {
+              if (changeset.get('isDirty')) {
+                changeset.execute()
+              }
+
+              await model.save()
```
czosel: try `changeset.save()`

anehx: Done!

PR #98, opened 2017-09-13
Update dependencies and resolve issues and deprecations

anehx:


app/styles/app.scss
```diff
@@ -239,3 +239,5 @@ body {
 .tt-suggestion:not(:hover) .inactive {
   background-color: rgb(230, 230, 230);
 }
+
+@import "ember-power-select";
```
czosel: why is this import not at the top?

package.json
```diff
@@ -10,8 +10,7 @@
   "scripts": {
     "build": "ember build",
     "start": "ember server --proxy http://localhost:8000",
-    "test": "ember test",
-    "docs": "ember ember-cli-yuidoc"
```
czosel: No more yuidoc? (Has this ever been used?)

anehx: No it hasn't - that's why..

app/styles/app.scss
anehx: This is duplicated.. The import is in `app/styles/ember-power-select-custom.scss`.

anehx: Fixed!

anehx:
Hooray!

czosel:
LGTM! :100:

PR #99, opened 2017-09-14
Add docker build

hairmare:
Build a docker container based on node that builds the final dist
and exposes it on port 80 using a basic nginx configuration.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,28 @@
+FROM node:6
+
+RUN npm install -g \
+      bower \
+ && apt-get update \
+ && apt-get install -y --no-install-recommends \
+      nginx \
+ && rm -rf /var/lib/apt/lists/* \
+ && rm /var/www/html/index.nginx-debian.html \
+ && ln -sf /dev/stdout /var/log/nginx/access.log \
+ && ln -sf /dev/stderr /var/log/nginx/error.log
+
+COPY . /usr/src/app
+
+WORKDIR /usr/src/app
+
+RUN npm install \
+ && bower install --allow-root \
+ && npm run-script build -- --environment=production \
+ && cp -r /usr/src/app/dist/* /var/www/html/ \
+ && mv contrib/nginx.conf /etc/nginx/sites-enabled/default
+
+WORKDIR /var/www/html
+
+EXPOSE 80
+
+ENTRYPOINT []
+CMD ["nginx", "-g", "daemon off;"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,8 @@
 FROM node:6
 
-RUN npm install -g \
+RUN yarn global add \
+      npm@5 \
+ && npm install -g \
       bower \
  && apt-get update \
  && apt-get install -y --no-install-recommends \
@@ -16,7 +18,7 @@ WORKDIR /usr/src/app
 
 RUN npm install \
  && bower install --allow-root \
- && npm run-script build -- --environment=production \
+ && npm run build -- --environment=production \
  && cp -r /usr/src/app/dist/* /var/www/html/ \
  && mv contrib/nginx.conf /etc/nginx/sites-enabled/default
 
--- a/contrib/nginx.conf
+++ b/contrib/nginx.conf
@@ -1,0 +1,17 @@
+server {
+    listen 80 default_server;
+    listen [::]:80 default_server;
+
+    root /var/www/html;
+
+    server_name _;
+
+    location /index.html {
+        expires epoch;
+        add_header Cache-Control 'private, no-store, must-revalidate, max-age=0';
+    }
+
+    location / {
+        try_files $uri $uri/ /index.html?/$request_uri;
+    }
+}
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+node_modules/
+bower_components/
+.git/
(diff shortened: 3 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM node:7
```
sliverc: I think we use node 6 for now

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+      yarn \
+      ember-cli \
+      bower \
+ && apt-get update \
+ && apt-get install -y --no-install-recommends \
+      nginx \
+ && rm -rf /var/lib/apt/lists/* \
+ && rm /var/www/html/index.nginx-debian.html \
+ && ln -sf /dev/stdout /var/log/nginx/access.log \
+ && ln -sf /dev/stderr /var/log/nginx/error.log
```
sliverc: what is the reason to have npm install and apt-get in one command? Cannot they be split into two RUN has they do not depend on each other?

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+COPY . /usr/src/app
+
+WORKDIR /usr/src/app
+
+RUN yarn install \
```
sliverc: I think #98 changes this again that yarn is not used anymore... may we need to wait for that PR to be merged.

contrib/nginx.conf
```diff
@@ -0,0 +1,16 @@
...
+	root /var/www/html;
+
+	index index.html;
+
+	server_name _;
+
+	location / {
+		# First attempt to serve request as file, then
+		# as directory, then fall back to displaying a 404.
+		try_files $uri /index.html =404;
```
sliverc: additionally we will also need to disable caching of index.html and add some headers for proper testing:
<pre>
        expires epoch;
        add_header Cache-Control 'private, no-store, must-revalidate, max-age=0';
        add_header Strict-Transport-Security max-age=63072000;
        add_header X-Frame-Options DENY;
        add_header X-Content-Type-Options nosniff;
        add_header Content-Security-Policy "default-src 'self'; img-src 'self' data:; style-src 'self' 'unsafe-inline'; object-src 'none'";
        add_header X-XSS-Protection "1; mode=block";
</pre>

I guess the  Strict-Transport-Security should be set by ingress later on... not sure how to accomplish it though.

Dockerfile
hairmare: The reasoning behind this was having all the installed stuff in one docker layer as per [the docs](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#minimize-the-number-of-layers). I don't think doing this in one `RUN` affects the readability of the Dockerfile.

Dockerfile
hairmare: Yeah, I missed that one and will rebase.

contrib/nginx.conf
hairmare: Why would we want to disable caching?

Dockerfile
sliverc: fair point... potentially we need to have a look at this again in timed-backend as well.

contrib/nginx.conf
hairmare: Got it, the index.html is really small and caching it would make updating the frontend harder than it should be.

Dockerfile
hairmare: Done

Dockerfile
hairmare: I just pushed the rebase (and also used amend since I had to rebase anyway)

hairmare:
@anehx Do you have any idea why travis is failing? I don't think I touched anything that reasonably explains the fail 😕

Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM node:6
+
+RUN npm install -g \
+      bower \
```
anehx: Add `npm@5` here to be sure we use npm version 5. Default is 3.x.x AFAIK.

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+ && ln -sf /dev/stdout /var/log/nginx/access.log \
+ && ln -sf /dev/stderr /var/log/nginx/error.log
+
+COPY . /usr/src/app
+
+WORKDIR /usr/src/app
+
+RUN npm install \
+ && bower install --allow-root \
+ && npm run-script build -- --environment=production \
```
anehx: `npm run build -- --environment=production` should be fine.. No need for `npm run-script`, nobody uses this anymore.

anehx:
@hairmare That one test fails sometimes and sometimes not.. Gotta look into this. Just restart the build and you'll be fine.

Dockerfile
hairmare: [this](https://stackoverflow.com/a/44280995) is fun. As the stackoverflow controbutor puts it: "¯\_(ツ)_/¯".

Basically since npm<5 is broken I'll need to use yarn to update npm@5. Luckily yarn is already in the node docker image.

hairmare: Done in 0dcb81aaa2cccd0d4f9a94fbf4ee7d877d5d2cd4

Dockerfile
hairmare: Also in 0dcb81aaa2cccd0d4f9a94fbf4ee7d877d5d2cd4

PR #100, opened 2017-09-14
Mobile optimization

anehx:
* Prevent mobile users from accessing `/reschedule` and `/analysis`
* Some responsive grid changes


app/mixins/desktop-only-route.js
```diff
@@ -0,0 +1,37 @@
...
+   * @method beforeModel
+   * @param {Ember.Transition} transition The transition
+   * @return {*} The return value of the parent function or a new transition
+   * @public
+   */
+  beforeModel(transition) {
+    if (
+      Object.values(
+        this.getProperties('media.isMd', 'media.isLg', 'media.isXl')
+      ).some(i => i)
```
czosel: This looks a bit funny, but i can't think of a good alternative... What came to mind was `.some(Boolean)` or `.some(i => !!i)` but that's also not any easier on the eye.

czosel: Well i guess most readable would be the boring way: `this.get('media.isMd') || this.get('media. ... `

app/reschedule/route.js
```diff
@@ -65,25 +70,26 @@ export default Route.extend(StaffRouteMixin, ReportFilterRouteMixin, {
...
+          /* istanbul ignore next */
+          this.get('notify').error('Error while verifying the page')
+        } finally {
+          this.send('finished')
+        }
       }
     }
   }
-})
+)
```
czosel: Prettier updates .... ^^

czosel: Ah, not even. I thought this was because of the new way prettier is laying out chained function calls. nvmd

anehx: Wait a second.. That's weird..

app/mixins/desktop-only-route.js
anehx: Yeah you're right..

app/reschedule/route.js
anehx: OHGODNO that's ugly..

anehx:
```javascript
const BaseRoute = Route.extend(
  StaffRouteMixin,
  ReportFilterRouteMixin,
  DesktopOnlyRouteMixin,
  {}
)

/**
 * Route for filtering and rescheduling reports
 *
 * @class RescheduleRoute
 * @extends Ember.Route
 * @using StaffRouteMixin
 * @using ReportFilterRouteMixin
 * @using DesktopOnlyRouteMixin
 * @public
 */
export default BaseRoute.extend({ ... })
```
Worse or better?

czosel: We're using prettier - there is no more "ugly" or "beautiful" just "prettier-conformant" or not :stuck_out_tongue_closed_eyes:

czosel: better!

PR #104, opened 2017-09-18
Minor absence modal fixes

anehx:
* Rollback the modal data after closing it
* Set the center of the calendar to the selected date
* Preselect the selected date in the new absence calendar

### DIFF
--- a/tests/acceptance/index-test.js
+++ b/tests/acceptance/index-test.js
@@ -140,7 +140,7 @@ describe('Acceptance | index', function() {
     expect(document.title).to.match(/\d{2}:\d{2}:\d{2} \(Unknown Task\)/)
   })
 
-  it('can add an absence for multiple days', async function() {
+  it('can add an absence for multiple days and current day is preselected', async function() {
     server.loadFixtures('absence-types')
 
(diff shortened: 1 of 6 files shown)


tests/acceptance/index-test.js
```diff
@@ -219,4 +218,24 @@ describe('Acceptance | index', function() {
     expect(find('[data-test-weekly-overview-day=29].holiday')).to.have.length(1)
     expect(find('[data-test-weekly-overview-day=28].holiday')).to.have.length(0)
   })
+
+  it('rollbacks the absence modal', async function() {
+    await visit('/?day=2017-06-29')
+
+    await click('[data-test-add-absence]')
+
+    await click('[data-date=2017-06-30]')
```
czosel: I was a little confused reading this. Wouldn't it be more logical if you'd move this await below the first expect call?
Also, I would have found it more logical to expect the 29th to be selected in both cases, instead of expecting the 30th not to be selected in the last expect call.

anehx: Seems fair.. Fixed!

PR #108, opened 2017-09-19
Fix hidden form rows in firefox

anehx:
* Use divs styled as table elements for form tables
* Change all values on change not on input


sliverc:
LGTM now.

PR #115, opened 2017-10-05
Handle activities over multiple days

anehx:
* Activity blocks only save times instead of datetimes
* Split activities which overlap the day into multiple activities on stop
* Ignore active activities when generating the timesheet

Needs to be merged together with the [backend PR](https://github.com/adfinis-sygroup/timed-backend/pull/87)


sliverc:
LGTM now.

PR #123, opened 2017-10-20
Statistics

anehx:


app/statistics/template.hbs
```diff
@@ -0,0 +1,137 @@
...
+{{fromDate}}
+<div>
+  {{#task-selection
+    history         = false
+    archived        = true
+    on-set-customer = (action 'setModelFilter' 'customer')
+    on-set-project  = (action 'setModelFilter' 'project')
+    on-set-task     = (action 'setModelFilter' 'task')
+    initial         = (hash
+      customer      = selectedCustomer
```
kaldras: Is this indent intended?

anehx: Yeah. `initial` is an object created with the `hash` helper and `customer` is a property of `initial`. But `{{fromDate}}` on line 3 is a leftover I just discovered ^^

kaldras: ah yes, right ^^ Everything fine then

PR #132, opened 2017-10-31
User profile credits tab

anehx:
Based off #131

### DIFF
--- a/app/users/edit/credits/controller.js
+++ b/app/users/edit/credits/controller.js
@@ -53,7 +53,7 @@ export default Controller.extend(UsersEditCreditsQueryParams.Mixin, {
       user: this.get('model.id'),
       include: 'absence_type',
       ordering: '-date',
-      ...(year ? { year } : [])
+      ...(year ? { year } : {})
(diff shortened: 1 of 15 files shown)


app/users/edit/credits/controller.js
```diff
@@ -0,0 +1,69 @@
...
+  },
+
+  absenceCredits: task(function*() {
+    let year = this.get('year')
+
+    return yield this.store.query('absence-credit', {
+      user: this.get('model.id'),
+      include: 'absence_type',
+      ordering: '-date',
+      ...(year ? { year } : [])
```
czosel: This looks crazy, but i guess it makes sense :smile: One question though: Can year be undefined when you define a defaultValue?

anehx: It can be whatever you set it. In this case, I set it to explicitly to `null` which is why I do this thingy there..

anehx: Actually this should be `(year ? { year } : {})`, which would make it more readable!

PR #139, opened 2017-11-07
Editable credits and year transfer

anehx:
* Add edit / new routes for overtime and absence credits
* Add transfer action for superusers in the last year

**TODO**
- [x] Deprecate `no-permission` template and transform it into a component
- [x] Show loading state on action buttons


anehx:
Now ready for review! @czosel @kaldras

PR #142, opened 2017-11-07
Update vertical collection to 1.0.0 beta 9

anehx:


anehx:
Will be included in #139 since this one contains dependency changes too..

PR #151, opened 2017-11-09
Drop reschedule and analysis page

anehx:
* Delete routes
* Delete tests
* Delete mixins only for those routes
* Delete components only for those routes
* Delete all staff logic since we don't do that anymore
* Delete all desktop only logic


anehx:
@kaldras Please do not merge this yet. I want to squash this one..

PR #167, opened 2017-11-16
Analysis page with infinite scrolling and async export

anehx:
* Add new analysis page with infinite scrolling table (requires `ember-scrollable` and `ember-in-viewport`)
* Add possibility of async download with `downloadjs`

TODO:

- [x] Add tests for download
- [x] Add tests for infinite scrolling


app/analysis/controller.js
```diff
@@ -0,0 +1,310 @@
...
+        /* istanbul ignore next */
+        if (!res.ok) {
+          throw new Error(res.statusText)
+        }
+
+        let file = yield res.blob()
+
+        let filename =
+          res.headers.map['content-disposition']
+            .match(/filename[^;=\n]*=((['"]).*?\2|[^;\n]*)/g)[0]
```
czosel: This line might deserve a comment

anehx: Agree.

app/analysis/template.hbs
```diff
@@ -0,0 +1,255 @@
...
+            <col>
+            <col>
+            <col>
+            <col>
+            <col>
+            <col>
+            <col>
+            <col>
+            <col>
+          </colgroup>
```
czosel: For what do you need the colgroup?

anehx: `colgroup` and `col` is needed for defining column widths in a table with `table-layout: fixed;`

czosel: alright, thanks!

app/analysis/controller.js
anehx: Done!

PR #175, opened 2017-11-24
chore(package): update eslint-plugin-ember to version 5.0.1

anehx:
Closes #173


app/components/task-selection/component.js
```diff
@@ -237,13 +240,15 @@ export default Component.extend({
       return value
     }
   },
+  /* eslint-enable ember/avoid-leaking-state-in-ember-objects */
```
czosel: Hmm, the setters in computed properties are really not too nice, right? Did you consider refactoring this?

anehx:
@czosel We discussed this right? So could you approve this one?

app/components/task-selection/component.js
anehx: BTW: Reason for this ignore thingy: https://github.com/ember-cli/eslint-plugin-ember/issues/42

PR #177, opened 2017-11-28
Improve handling of negative durations for the picker and display functions

anehx:
For better understanding, see the new tests which show the bugs we had before.


app/utils/humanize-duration.js
```diff
@@ -21,15 +21,17 @@ export default function humanizeDuration(duration, seconds = false) {
     return seconds ? '0h 0m 0s' : '0h 0m'
   }
 
+  let prefix = duration.asMinutes() < 0 ? '-' : ''
```
czosel: nitpick: wouldn't `asSeconds` be more correct? (What are negative durations used for, anyway?)

app/components/sy-durationpicker/component.js
```diff
@@ -70,6 +72,26 @@ export default SyTimepickerComponent.extend({
...
+   * @return {moment.duration} The resulted duration
+   * @private
+   */
+  _getDurationFromHoursAndMinutes(h, m) {
+    let negative = [h, m].some(n => n < 0)
+
+    h = abs(h) * (negative ? -1 : 1)
+    m = abs(m) * (negative ? -1 : 1)
+
+    return moment.duration({ h, m })
```
czosel: In the test, you're trying a negative value for `h` and a positive one for `m`. Is the only case that makes sense? If yes, i'd also implement it explicitly like that, and not generically like above, because like this you have 
```
h=1, m=-30 => -1:30
h=-1, m=-30 => -1:30
```
which might be a little confusing.

app/utils/humanize-duration.js
anehx: Maybe `.milliseconds()` would be best.. Negative durations are used e.g. for negative overtime-credits

app/components/sy-durationpicker/component.js
anehx: I see your point. Maybe we should extract negative as a parameter and pass the hours and minutes as positive numbers. What do you think?

anehx: ATM we extract hours and minutes from a string by `let [h, m] = str.split(':').map(n => parseInt(n))` which results in only negative hours but not minutes. But we also have the use case where we pass negative minutes to add (when using the arrow keys). So extracting a parameter whether the duration is negative or not might be the best solution.

czosel: As discussed: extra parameter is fine!

app/utils/humanize-duration.js
czosel: As discussed: leaving out `asMinutes` and relying on default bahivour of moment should work, too. If not, it might be `asMilliseconds`.

anehx:
@czosel discussed stuff is done!

anehx:
@czosel Any comments on this?

PR #182, opened 2017-12-05
Make rows in analysis selectable

anehx:
@sliverc Please check if the permission checking is done right (in `app/abilities/report.js`)


app/abilities/report.js
```diff
@@ -0,0 +1,23 @@
...
+  canEdit(
+    currentUserId = null,
+    isSuperuser = false,
+    userId = null,
+    supervisors = [],
+    reviewers = []
+  ) {
+    return (
+      isSuperuser ||
+      userId === currentUserId ||
```
anehx: Only the superuser can edit verified reports!

PR #184, opened 2017-12-08
Analysis bulk edit

anehx:
* Add indeterminate state to `sy-checkbox`
* Add bulk edit mask for analysis

Based on #182 which needs to be merged first..


app/analysis/edit/controller.js
```diff
@@ -0,0 +1,153 @@
...
+      )
+      let value = get(params, key)
+
+      return key === 'type'
+        ? parsed
+        : {
+            ...parsed,
+            [underscore(key)]: serialize ? serialize(value) : value
+          }
+    }, {})
```
czosel: This seems to be pretty much that same as in the statistics controller. Since the method is pretty high-level it might be worth to

* extract it
* add a docstring

app/analysis/edit/controller.js
```diff
@@ -0,0 +1,153 @@
...
+  )
+}
+
+const filterUnchanged = (attributes, changes) => {
+  return Object.keys(attributes).reduce((obj, attr) => {
+    if (changes.map(({ key }) => dasherize(key)).includes(attr)) {
+      return { ...obj, [attr]: attributes[attr] }
+    }
+
+    return { ...obj }
```
czosel: Do you really need to clone `obj` here?

app/analysis/edit/controller.js
```diff
@@ -0,0 +1,153 @@
...
+    return id && this.store.peekRecord('task', id)
+  },
+
+  save: task(function*(changeset) {
+    try {
+      let params = prepareParams(this.get('allQueryParams'))
+
+      let queryString = Object.keys(params)
+        .map(k => `${k}=${params[k]}`)
+        .join('&')
```
czosel: use `toQueryString` in url util

app/analysis/edit/controller.js
```diff
@@ -0,0 +1,153 @@
...
+      }
+
+      yield this.get(
+        'ajax'
+      ).request(`/api/v1/reports/bulk?editable=1&${queryString}`, {
+        method: 'POST',
+        data: { data }
+      })
+
+      this.get('analysisController')._reset()
```
czosel: `_reset` looks like it's private - might be worth removing the underscore?

app/analysis/edit/controller.js
anehx: Done!

app/analysis/edit/controller.js
anehx: No. I simplified it.

app/analysis/edit/controller.js
anehx: Would be nice if I remembered my own utils ^^ Thanks!

app/analysis/edit/controller.js
anehx: It really is private now since I remove this very call here and added it to the index controllers `setup` hook.

PR #195, opened 2017-12-19
Allow hiding the filter sidebar

anehx:
* Use #-in-element for wormholing the sidebar
* Make sidebar off-canvas
* Hide some too large content on mobile


app/components/no-mobile-message/template.hbs
```diff
@@ -0,0 +1,12 @@
+<div>
+  {{fa-icon 'mobile'}}
+  <h3>Sorry, this page doesn't work on mobile!</h3>
+  <p>
+    The data on this page is taking up too much space for your currently used
+    device.
+  </p>
+  <p>
+    Try opening this page on a larger device. or take me back to
```
czosel: I'd suggest "Try opening this page on a larger device, or go back to the home page." - you're refering to the user as "you" in the sentence above.

czosel:
What do you think about hiding routes which are not available on mobile in the navigation instead of showing an error? Would people get confused why the menu they're used to is not available?

anehx:
@czosel That's exactly what I'm doing..

czosel:
Ah, so the error message is only displayed when one ends up on such a route anyway - perfect!

anehx:
@czosel Yes exactly.. There are users which resize the window all the time :wink:

PR #198, opened 2017-12-19
Fix various minor issues

anehx:
* Improve style and position of the filter toggle button
* Show the selected filters on said button
* Make filter sidebar scrollable
* Fix applied filter count on statistics page
* Always set editable filter to true on bulk edit
* Reset selected row on filter changes and reset

### DIFF
--- a/app/analysis/edit/controller.js
+++ b/app/analysis/edit/controller.js
@@ -8,7 +8,8 @@ import computed from 'ember-computed-decorators'
 import { toQueryString } from 'timed/utils/url'
 import {
   underscoreQueryParams,
-  serializeParachuteQueryParams
+  serializeParachuteQueryParams,
+  filterQueryParams
 } from 'timed/utils/query-params'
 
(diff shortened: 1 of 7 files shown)


app/analysis/edit/controller.js
```diff
@@ -30,7 +31,10 @@ export const AnalysisEditQueryParams = AnalysisQueryParams.extend({
 const prepareParams = params =>
   cleanParams(
     underscoreQueryParams(
-      serializeParachuteQueryParams(params, AnalysisEditQueryParams)
+      serializeParachuteQueryParams(
```
czosel: Damn, this level of nesting would almost justify pulling in an FP lib to get `pipe` support ;)

anehx: Blargh *library*.. Waiting for this to be supported http://tc39.github.io/proposal-pipeline-operator/..

czosel: right - as discussed

PR #230, opened 2018-07-17
Make user-selection generic

Yelinz:
Make the user-selection component generic, for query parameters.


Yelinz:
It is done form my side.

PR #232, opened 2018-07-18
Add date preset buttons

Yelinz:
Add buttons for presetting the date filters.


app/components/date-buttons/component.js
```diff
@@ -0,0 +1,22 @@
+import Component from '@ember/component'
+import { A } from '@ember/array'
+import moment from 'moment'
+
+export default Component.extend({
+  choices: A(['this week', 'this month', 'this year']),
```
czosel: Do you really need an ember array here?

tests/integration/components/date-buttons/component-test.js
```diff
@@ -0,0 +1,38 @@
...
+import { click } from 'ember-native-dom-helpers'
+import moment from 'moment'
+
+describe('Integration | Component | date buttons', function() {
+  setupComponentTest('date-buttons', {
+    integration: true
+  })
+
+  it('changes the date', async function() {
+    let format = 'YYYY-MM-DD'
```
czosel: `const` instead of `let`

app/components/date-buttons/component.js
```diff
@@ -0,0 +1,21 @@
+import Component from '@ember/component'
+import moment from 'moment'
+
+export default Component.extend({
+  choices: ['this week', 'this month', 'this year'],
```
czosel: CI noticed that this has to be moved to init:
```
/home/travis/build/adfinis-sygroup/timed-frontend/app/components/date-buttons/component.js
  5:3  error  Only string, number, symbol, boolean, null, undefined, and function are allowed as default properties  ember/avoid-leaking-state-in-ember-objects
```

PR #233, opened 2018-07-19
Behaviour change for verify checkbox and edit all button 

Yelinz:
Instead of hiding the elements, they become disabled with a hover text.

### DIFF
--- a/app/analysis/edit/template.hbs
+++ b/app/analysis/edit/template.hbs
@@ -82,15 +82,18 @@
                   {{/sy-checkbox}}
                 {{/f.input}}
 
-                {{#if (or (eq reviewer session.data.user.id) session.data.user.isSuperuser)}}
-                  {{#f.input data-test-verified name='verified' as |fi|}}
-                    {{#sy-checkbox checked=fi.value on-change=fi.update}}
-                      Verified
-                      {{#if (eq model.verified null)}}{{not-identical-warning}}{{/if}}
-                      {{#unless (eq f.model.verified model.verified)}}{{changed-warning}}{{/unless}}
-                    {{/sy-checkbox}}
-                  {{/f.input}}
-                {{/if}}
+                {{#f.input data-test-verified name='verified' as |fi|}}
+                  {{#sy-checkbox
+                    checked=fi.value
+                    on-change=fi.update
+                    title=(unless (or (eq reviewer session.data.user.id) session.data.user.isSuperuser) 'Please set reviewer filter to yourself to verify reports.')
+                    disabled=(unless (or (eq reviewer session.data.user.id) session.data.user.isSuperuser) 'true')
+                  }}
+                    Verified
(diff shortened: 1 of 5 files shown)


app/analysis/edit/template.hbs
```diff
@@ -82,15 +82,18 @@
...
-                      {{#if (eq model.verified null)}}{{not-identical-warning}}{{/if}}
-                      {{#unless (eq f.model.verified model.verified)}}{{changed-warning}}{{/unless}}
-                    {{/sy-checkbox}}
-                  {{/f.input}}
-                {{/if}}
+                {{#f.input data-test-verified name='verified' as |fi|}}
+                  {{#sy-checkbox
+                    checked=fi.value
+                    on-change=fi.update
+                    title=(unless (or (eq reviewer session.data.user.id) session.data.user.isSuperuser) 'Please set reviewer filter to yourself to verify reports.')
```
czosel: Maybe better: "Please select yourself as "reviewer" to verify reports.

czosel: Also, it might be worth assigning `(unless (or (eq reviewer session.data.user.id) session.data.user.isSuperuser)` to a variable like `canVerify`.

app/analysis/index/template.hbs
```diff
@@ -179,9 +179,16 @@
...
     {{#with data.lastSuccessful.value as |reports|}}
       <div class="btn-toolbar btn-toolbar--right form-group">
-        {{#if editable}}
-          <button data-test-edit-all type="button" class="btn" {{action 'edit'}}>Edit all</button>
-        {{/if}}
+        <button
+          data-test-edit-all
+          type="button"
+          class="btn"
+          title={{unless editable 'Please set editable filter to enable edit all.'}}
```
czosel: Please enable the "editable" filter to modify all entries at once.

app/analysis/index/template.hbs
```diff
@@ -309,4 +316,4 @@
       </div>
     {{/with}}
   {{/if}}
-{{/if}}
\ No newline at end of file
+{{/if}}
```
czosel: Where is the extra `if` coming from?

czosel: Right - I'm just confused about the diff: It looks like you added a closing `if` without adding an opening one?

Yelinz: Vim added an EOL, which the previous editor didnt.

czosel: Oh right. We should add https://eslint.org/docs/rules/eol-last and run `--fix` to enforce consistent style.

czosel: Can you do that in a separate MR?

Yelinz: So add EOLs or remove them?

czosel: It seems that the Unix default is to add them (just like the Vim and the ESlint default), so i'd say let's add them :wink:

PR #246, opened 2018-08-06
Feature analysis persistent location

Yelinz:
The location to where you scrolled to gets saved, so after editing, you don't need to scroll all the way back.


app/analysis/index/controller.js
```diff
@@ -162,6 +162,8 @@ const AnalysisController = Controller.extend(AnalysisQueryParams.Mixin, {
 
   @oneWay('session.data.authenticated.token') jwt: null,
 
+  _scrollOffset: 0,
```
anehx: Reset this property to zero on reset (line 191)

Yelinz: Done

PR #250, opened 2018-08-22
Redesign activities

Yelinz:


app/index/activities/controller.js
```diff
@@ -19,10 +20,13 @@ export default Controller.extend({
...
-   */
-  @computed()
-  _allActivities() {
+  _allActivities: computed(function() {
     return this.store.peekAll('activity')
+  }),
+   */
+  init() {
+    this._super(...arguments)
+    this.set('_allActivities', this.store.peekAll('activity'))
```
anehx: Why are you doing it that way?

app/index/activities/edit/route.js
```diff
@@ -75,19 +74,18 @@ export default Route.extend(RouteAutostartTourMixin, {
     this._super(...arguments)
 
     let activity = new Changeset(
-      model,
+      this.store.findRecord('activity', model.belongsTo('activity').id(), {
+        include: 'blocks'
+      }),
```
anehx: What exactly are you trying to do here?

Yelinz: This is an old version, to implement the new design, now changed.

app/index/activities/controller.js
Yelinz: Instead of having an empty computed, which also only runs once when it gets called the first time.
I put it into init for the same effect.

app/components/task-selection/component.js
```diff
@@ -270,28 +262,26 @@ export default Component.extend({
         })
       }
 
-      later(this, () => {
+      later(this, async () => {
```
anehx: Why `async`?

Yelinz: Was for testing purposes, removed it

anehx:
@Yelinz as discussed, this PR consists of way to many changes which should be splitted. PRs for updates or deprecation fixes can't be in the same as logic changes.

PR #268, opened 2018-10-25
Implementation of the new component review-warning

aziiee:


app/components/review-warning/component.js
```diff
@@ -0,0 +1,34 @@
...
+        to_date: this.endOfMonth.format('YYYY-MM-DD'),
+        reviewer: this.get('session.data.user.id'),
+        editable: 1
+      })
+      .then(reports => {
+        // Hide review warning button if no reports found
+        if (reports.get('length') > 0) {
+          this.set('isVisible', true)
+        }
+      })
```
czosel: Could we use async/await instead of promises here?

app/components/review-warning/component.js
```diff
@@ -0,0 +1,34 @@
...
+export default Component.extend({
+  isVisible: false,
+  session: service('session'),
+  store: service('store'),
+  startOfMonth: moment()
+    .startOf('month')
+    .subtract(1, 'months'),
+  endOfMonth: moment()
+    .endOf('month')
+    .subtract(1, 'months'),
```
anehx: This should actually be called on `didInsertElement` rather than at the time the component definition is called

app/components/review-warning/template.hbs
```diff
@@ -0,0 +1,10 @@
+<li class="nav-top-list nav-top-list--right">
+      {{#link-to 'analysis' (query-params
+      reviewer=session.data.user.id
+      editable=1
+      fromDate=startOfMonth
+      toDate=endOfMonth
+      )data-test-review-warning}}
+        {{fa-icon 'exclamation-triangle'}} Review warning
+      {{/link-to}}
```
anehx: Proper formatting please

app/components/review-warning/component.js
aziiee: Could you review again?

app/components/review-warning/component.js
aziiee: Updated that but I have used `init` instead of `didInsertElement`. In my opinion using `didInsertElement` only makes sense if reports are queried in `didInsertElement` as well because `startOfMonth` and `endOfMonth` need to initialized before querying reports.
What is your opinion on that?

tests/acceptance/index-test.js
```diff
@@ -268,8 +268,10 @@ describe('Acceptance | index', function() {
     await visit('/')
     expect(find('[data-test-review-warning]')).to.have.length(1)
     await click('[data-test-review-warning]')
-    expect(currentURL()).to.equal(`
+
+    let expectedUrl = `
```
czosel: If you're not mutating the variable, we usually prefer to use `const` over `let`.

app/components/review-warning/component.js
czosel: @anehx Just discussed with @aziiee, and the `init` hook seems fine to both of us. Feel free to correct us though :wink:

czosel:
@aziiee Tests still seem to be failing though.

aziiee:
Closing, since this has been implemented in #380

PR #281, opened 2018-11-13
Fix persistent scroll in analysis

Yelinz:


app/analysis/edit/controller.js
```diff
@@ -127,6 +127,18 @@ export default Controller.extend(AnalysisEditQueryParams.Mixin, {
         'intersection.lastSuccessful.value.model'
       ).serialize()
 
+      params.id.split(',').forEach(async id => {
+        if (id) {
```
czosel: Maybe use `filter` instead of `if` inside `forEach`?

app/analysis/index/controller.js
```diff
@@ -193,6 +193,8 @@ const AnalysisController = Controller.extend(AnalysisQueryParams.Mixin, {
 
   _scrollOffset: 0,
 
+  dataIntegrity: true,
```
czosel: As discussed: This deserves a comment.

app/analysis/edit/controller.js
Yelinz: Added ```without('')``` the problem was for an empty string it would put an empty string into the array.

app/analysis/index/controller.js
```diff
@@ -205,6 +207,11 @@ const AnalysisController = Controller.extend(AnalysisQueryParams.Mixin, {
 
     if (!this.get('skipResetOnSetup')) {
       this._reset()
+    } else {
+      this.setProperties({
+        selectedReportIds: A(),
```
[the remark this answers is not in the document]
Yelinz: Native suffices, it anyways gets replaced when an report gets selected

app/analysis/index/controller.js
Yelinz: Added an comment, check if it makes it more understandable

czosel: :+1:

app/analysis/edit/controller.js
czosel: Awesome, didn't know about `without`!

PR #288, opened 2018-12-04
Fix analysis, after update it doesnt set verifier

Yelinz:
Also add highlighting of past edited reports.

### DIFF
--- a/app/analysis/edit/controller.js
+++ b/app/analysis/edit/controller.js
@@ -134,6 +134,10 @@ export default Controller.extend(AnalysisEditQueryParams.Mixin, {
           let report = await this.store.peekRecord('report', id)
           if (report) {
             changeset.get('changes').forEach(obj => {
+              if (obj.key === 'verified') {
+                obj.key = 'verifiedBy'
+                obj.value = obj.value ? this.get('session.data.user') : null
+              }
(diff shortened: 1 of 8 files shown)


mirage/factories/task.js
```diff
@@ -5,7 +5,7 @@ import { randomDuration } from '../helpers/duration'
 export default Factory.extend({
   name: () => capitalize(faker.hacker.ingverb()),
   estimatedTime: () => randomDuration(),
-  // project: association()
+  // project: association(),
```
kaldras: Is this comment really needed?

app/analysis/edit/controller.js
```diff
@@ -134,7 +134,16 @@ export default Controller.extend(AnalysisEditQueryParams.Mixin, {
           let report = await this.store.peekRecord('report', id)
           if (report) {
             changeset.get('changes').forEach(obj => {
-              report.set(obj.key, obj.value)
+              let key = obj.key
+              let val = obj.value
+              if (key === 'verified' && val) {
```
kaldras: As discussed, logic is fine but let us shorten this code part. ;)

mirage/factories/task.js
Yelinz: It isnt

PR #336, opened 2019-11-20
Add additional filter options for reports

aziiee:
This will allow to filter reports by last week, last month and last
year in Analysis and Statistics.

### DIFF
--- a/app/statistics/template.hbs
+++ b/app/statistics/template.hbs
@@ -107,7 +107,10 @@
             on-change = (action (mut toDate))
           }}
         {{/fs.label}}
-        {{date-buttons presetDate=(action (mut fromDate))}}
+        {{date-buttons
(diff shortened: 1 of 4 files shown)


app/statistics/template.hbs
```diff
@@ -107,7 +107,10 @@
             on-change = (action (mut toDate))
           }}
         {{/fs.label}}
-        {{date-buttons presetDate=(action (mut fromDate))}}
+        {{date-buttons
+            presetFromDate=(action (mut fromDate))
+            presetToDate=(action (mut toDate))
```
anehx: If I understand correctly, `presetToDate` is an action. However the naming suggests otherwise. Please rename this so the type of the argument is clear, e.g `onUpdateToDate`

PR #365, opened 2020-02-07
chore(deps): adapt to backend jwt changes

sbor23:
Rename data type for JWT login requests as needed by backend.
See: https://github.com/adfinis-sygroup/timed-backend/pull/502


sbor23:
This needs some more work. Based on the ember upgrade branch this throws an error, even though the backend correctly issues a token.

PR #367, opened 2020-02-12
feat: add Adsy ESLint config and update code

fkm:


fkm:
> Looks about right, cant really say with over 300 files and mostly only quotation and semicolon changes

The only real change is the ESLint config :smiley: The rest are changes needed from the config change.

PR #380, opened 2020-02-27
feat(indicator): Show review warning for unverified reports

aziiee:
The review warning is displayed to reviewers that haven't verified reports for last month or earlier yet.

### DIFF
--- a/app/services/unverified-reports.js
+++ b/app/services/unverified-reports.js
@@ -18,6 +18,7 @@ export default Service.extend({
   store: service(),
   session: service(),
   notify: service(),
+  amountReports: 0,
 
   hasReports: computed("amountReports", function() {
     return this.get("amountReports") > 0;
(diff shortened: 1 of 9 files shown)


app/services/unverified-reports.js
```diff
@@ -0,0 +1,60 @@
+import { computed } from "@ember/object";
+import Service, { inject as service } from "@ember/service";
+import moment from "moment";
+
+const INTERVAL_DELAY = 10 * 60000; // 10 Minutes
```
fkm: I usually go from the smallest unit to what I need. In this case:
`1000 * 60 * 10`

But this is just my personal preference. No need to change this!

app/services/unverified-reports.js
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * @class UnverifiedReportsService
+ * @extends Ember.Service
+ * @public
+ */
+export default Service.extend({
+  store: service(),
+  session: service(),
+  notify: service(),
+
```
fkm: I recommend adding the property "amountReports" with an initial value of zero.

This way you don't have weird errors/behavior if the property is read before the first polling is over.

PR #381, opened 2020-02-27
feat(responsibilities): show supervisee vacations

fkm:


fkm:
@aziiee I fixed the problem with the array. Using `firstObject` instead of `.[0].` did the trick.

PR #390, opened 2020-03-11
feat(projects): Allow to create and edit tasks

aziiee:


app/projects/route.js
```diff
@@ -0,0 +1,19 @@
...
+    this._super(...args);
+
+    if (this.user.isSuperuser || this.user.isReviewer) {
+      controller.set("hasAccess", true);
+      controller.fetchProjectsOfUser.perform();
+    } else {
+      controller.set("hasAccess", false);
+    }
+  }
+});
```
fkm: Why did you use a route?

app/projects/template.hbs
```diff
@@ -0,0 +1,115 @@
...
+  {{#if selectedProject}}
+    <h3>
+      <span class="header-left">
+        Tasks of {{selectedProject.name}}
+      </span>
+      <span class="header-right">
+        <button class="btn btn-primary" {{action (perform createTask)}} data-test-add-task>Add Task</button>
+      </span>
+    </h3>
+    <div style="height: 400px; margin-bottom: 20px">
```
fkm: See if you can find/create a CSS class you can use for that.

app/projects/route.js
aziiee: @fkm fixed used `init()` inside the controller instead

app/projects/template.hbs
aziiee: @fkm fixed

PR #391, opened 2020-03-12
chore: add docker setup

Yelinz:


docker-compose.override.yml
```diff
@@ -0,0 +1,34 @@
...
+    volumes:
+      - .:/myapp
+      - ./node_modules:/myapp/node_modules
+    tmpfs:
+      - /myapp/tmp
+    ports:
+      - 4200:4200
+      - 7020:7020
+      - 7357:7357
+    command: ember server --proxy=http://backend
```
anehx: The frontend container should be the same as in production. No one is developing using a container ATM.

docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+      - dbdata:/var/lib/postgresql/data
+    environment:
+      - POSTGRES_USER=timed
+      - POSTGRES_PASSWORD=timed
+
+  frontend:
+    build:
+      context: .
+    ports:
+      - 4000:80
```
anehx: Why 4000? Also, you need to document this

contrib/nginx.conf
```diff
@@ -2,16 +2,54 @@ server {
     listen 80 default_server;
     listen [::]:80 default_server;
 
+    resolver 127.0.0.11;
```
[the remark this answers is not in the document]
Yelinz: 11 is for the Docker internal DNS, couldn't get the rerouting to mailhog to work with 1

docker-compose.yml
Yelinz: Because if we use an override file, ports get merged and there would be a conflict having 4200 assigned twice

docker-compose.override.yml
Yelinz: Removed dev container

docker-compose.yml
Yelinz: Removed override, the port is 4200 now

PR #395, opened 2020-03-17
fix: budget label instead of estimate in progress tooltip

tk0e:
"Budget" seems more appropriate than "Estimate" and was also specified in the feature request for the progress tooltip


tongpu:
To fix the four failing tests could you please also replace `Estimate` with `Budget` in the respective tests: `tests/integration/components/progress-tooltip/component-test.js`

tk0e:
Fixed the tests with https://github.com/adfinis-sygroup/timed-frontend/pull/396

aziiee:
@tk0e Thanks for the fix!

PR #396, opened 2020-03-18
fix: budget instead of estimate in progress tooltip

tk0e:
Changed the label also in the tests. Change in the component was done with https://github.com/adfinis-sygroup/timed-frontend/pull/395


tongpu:
I changed the destination branch to `tkoe-patch-1` so that we can review both changes in #395 as soon as @aziiee approves the changes.

PR #401, opened 2020-03-25
Use Source Sans Pro instead of TheSansLT

aziiee:


aziiee:
@anehx Here is a summary of changes, since the diff is not much of a help:

- Overriden `font-family` rule to use `Source Sans Pro`, whereever `TheSansLT` was used
- Removed wrong implementation of `@font-face` in Adcssy

PR #408, opened 2020-03-26
fix(projects): Add customers option

aziiee:
There can be projects with the same name but under different
customers. Therefore the customers option is necessary.

### DIFF
--- a/app/projects/controller.js
+++ b/app/projects/controller.js
@@ -1,7 +1,8 @@
 import Controller from "@ember/controller";
-import { reads } from "@ember/object/computed";
+import { or, reads, uniqBy } from "@ember/object/computed";
 import { inject as service } from "@ember/service";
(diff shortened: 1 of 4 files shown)


app/projects/controller.js
```diff
@@ -11,21 +12,46 @@ export default Controller.extend({
...
+      } else {
+        projects = yield this.store.query("project", {
+          reviewer: this.get("user.id")
+        });
       }
-      return yield this.store.query("project", {
-        reviewer: this.get("user.id")
-      });
+
+      this.fetchCustomersOfProject.perform(projects);
```
fkm: Is it on purpose that you don't await/yield here?

aziiee: Yes, IMO it's not necessary since the project selection will be disabled unless there is a selected customer anyways. Feel free to correct me though

PR #416, opened 2020-04-02
fix(general): Increase readability

aziiee:
This will increase the readability with the new font througout the
application.


aziiee:
The only change is the increase of `font-weight` from `300` to `400`

PR #419, opened 2020-04-08
Fix(analysis): Set limit for reports export

aziiee:


app/analysis/index/controller.js
```diff
@@ -33,6 +33,8 @@ const serializeMoment = momentObject =>
 const deserializeMoment = momentString =>
   (momentString && moment(momentString, DATE_FORMAT)) || null;
 
+const EXPORT_LIMIT = 50000;
```
aziiee, line 36: @sbor23 Please confirm the export limit.

sbor23:
please wait before you merge this. At the moment 100'000 would work but I'm looking to into solving this problem completely using annotation. So in the best case we don't need this at all.

Anyway this number shouldn't be hardcoded but made configurable, also in the frontend if possible.
System performance differs a lot and there would be no guarantee that something that works locally would also work the test/prod server.

sbor23:
See https://github.com/adfinis-sygroup/timed-backend/pull/541 for the backend part which is configurable.

app/analysis/index/template.hbs
```diff
@@ -305,6 +305,8 @@
             type="button"
             data-download-count="{{download.performCount}}"
             class="btn btn-primary {{if (and (eq download.last.url link.url) (eq download.last.params link.params) download.isRunning) 'loading'}}"
+            disabled={{exportLimitExceeded}}
+            title={{if exportLimitExceeded "Please use filters to reduce the amount of reports to export."}}
```
anehx, line 309: Include the export maximum here so the user knows it

aziiee: @anehx done

app/analysis/index/controller.js
```diff
@@ -184,8 +184,20 @@ const AnalysisController = Controller.extend(AnalysisQueryParams.Mixin, {
...
   ),
 
+  exportLimitExceeded: computed("totalItems", function() {
+    return this.get("totalItems") > this.get("exportLimit");
+  }),
+
+  exportLimitMessage: computed("exportLimit", function() {
+    return `The export limit is ${this.get(
+      "exportLimit"
+    )}. Please use filters to reduce the amount of reports.`;
```
anehx, line 194:
```suggestion
    return `The export limit is ${this.exportLimit}. Please use filters to reduce the amount of reports.`;
```

`.get` is not necessary nowadays

aziiee: should be okay now

PR #422, opened 2020-04-14
chore(deps): Update ember-scrollable

aziiee:
It's not described in their changelog but bumping the version will fix the issue where an empty scrollbar is displayed at the bottom after visiting a page that uses this addon.

Tested locally on `/analysis` and `/projects`


aziiee:
Wondering why it didn't install `v1.0.2` initially since I haven't touched `package.json`

PR #423, opened 2020-04-16
chore(release): v0.13.1

aziiee:
Starting with `v0.13.1`, the first bugfix release in `v0.13` series.


aziiee:
> The commit message is not valid, we use the conventional commit standard..
> 
> It should be something like `chore(release): v0.13.1`

@anehx fixed

PR #425, opened 2020-04-21
fix(review-warning): Reset query parameters

aziiee:


app/components/report-review-warning/template.hbs
```diff
@@ -2,10 +2,19 @@
...
                  editable=1
                  verified=0
+                 customer=null
+                 project=null
+                 task=null
+                 user=null
+                 billingType=null
+                 costCenter=null
+                 review=""
+                 notBillable=""
```
aziiee, line 17: @anehx I'm not sure if this is the right approach to my problem since I am resetting the query params manually. Do you maybe know a better way or can we leave it like this?

anehx: Hm looks weird but I don't have a better idea

PR #445, opened 2020-06-04
feat(auth): Replace current auth flow with OIDC

aziiee:


app/protected/route.js
```diff
@@ -23,24 +23,28 @@ export default Route.extend(AuthenticatedRouteMixin, {
...
+  ajax: service(),
 
   async model() {
-    const id = this.get("session.data.authenticated.user_id");
+    const user = await this.ajax.request("/api/v1/users/me", {
+      method: "GET"
+    });
 
-    const user = await this.store.findRecord("user", id, {
+    const usermodel = await this.store.findRecord("user", user.data.id, {
```
anehx, line 34: Don't you have that ID in the session? I don't think the `/api/v1/me` call is necessary

config/environment.js
```diff
@@ -94,6 +102,8 @@ module.exports = function(environment) {
 
   // eslint-disable-next-line no-empty
   if (environment === "production") {
+    ENV["ember-simple-auth-oidc"].host =
+      "https://timed.adfinis-sygroup.ch/auth/realms/timed/protocol/openid-connect";
```
anehx, line 106: Shouldn't this be our SSO?

aziiee:
@anehx I am doing the `/api/v1/me` request because the user IDs in keycloak don't match with the User IDs in the API

PR #473, opened 2020-08-21
Revert "fix(auth): Redirect to SSO if authentication fails"

aziiee:
Reverts adfinis-sygroup/timed-frontend#471

There is no need to redirect to the login page manually as `ember-simple-auth` does that anyway if a request is made with an invalid token.


aziiee:
\cc @michaelimfeld

PR #484, opened 2020-10-02
feat: introduce code owners file

winged:
Let's see how this works out. This here is more of an experiment than
a production setting. If it gains traction, we'll add it to more projects
in the future

See also adfinis-sygroup/timed-backend#634


winged:
Updated team reference and renamed teams. Nested teams are not referenced by nested names

hairmare:
I prepared repo permissions for @adfinis-sygroup/dev-frontend so this should work after we merge it.

hairmare (approved):

PR #560, opened 2021-06-21
chore: drop CODEOWNERS file

czosel:
I think it's more reasonable to leave notifications up to personal configuration. A large part of the frontend team is not working actively on this project, and the code ownership generates unnecessary notifications.


winged (approved):
Good call

PR #562, opened 2021-06-22
feat: visualise  customer visible projects

trowik:
add an icon to the comment field for customer visible projects
customer_visible_icon_2
customer_visible_icon

### DIFF
--- a/app/components/customer-visible-icon/component.js
+++ b/app/components/customer-visible-icon/component.js
@@ -1,0 +1,6 @@
+import Component from "@ember/component";
+
+export default Component.extend({
+  attributeBindings: ["title"],
+  title: "This project's comments are visible to the customer"
+});
--- a/app/components/customer-visible-icon/template.hbs
+++ b/app/components/customer-visible-icon/template.hbs
@@ -1,0 +1,1 @@
+<FaIcon @icon="eye" />
--- a/app/components/report-row/template.hbs
+++ b/app/components/report-row/template.hbs
@@ -15,6 +15,9 @@
     onchange={{action (mut changeset.comment) value='target.value'}}
     data-test-report-comment
   >
+  {{#if changeset.task.project.customerVisible }}
+      <CustomerVisibleIcon class="customer-visible-icon" />
+  {{/if}}
 </div>
 <div class="form-list-cell form-group">
   {{sy-durationpicker-day data-test-report-duration disabled=(not editable) value=changeset.duration on-change=(action (mut changeset.duration))}}</div>
--- a/app/components/tracking-bar/template.hbs
+++ b/app/components/tracking-bar/template.hbs
@@ -24,6 +24,9 @@ as |t|}}
         value       = tracking.activity.comment
         disabled    = tracking.activity.active
       }}
+      {{#if tracking.activity.task.project.customerVisible }}
+        <CustomerVisibleIcon class="customer-visible-icon" />
+      {{/if}}
     </div>
     <div class="form-group">
       {{record-button
--- a/app/index/activities/edit/template.hbs
+++ b/app/index/activities/edit/template.hbs
@@ -18,6 +18,9 @@
           value={{changeset.comment}}
           onchange={{action (mut changeset.comment) value='target.value'}}
         >
+        {{#if changeset.task.project.customerVisible }}
+          <CustomerVisibleIcon class="customer-visible-icon" />
+        {{/if}}
       </div>
       <div class="form-group checkboxes">
         {{sy-checkbox
--- a/app/index/activities/template.hbs
+++ b/app/index/activities/template.hbs
@@ -43,7 +43,12 @@
                   <strong>Unknown task</strong>
                 {{/if}}
               </td>
-              <td title={{activity.comment}}><div>{{activity.comment}}</div></td>
+              <td title={{activity.comment}}>
+                <div>{{activity.comment}}</div>
+                {{#if activity.task.project.customerVisible }}
+                  <CustomerVisibleIcon class="activity-customer-visible-icon" />
+                {{/if}}
+              </td>
               <td>
                 <div>{{sy-checkmark checked=activity.review}} Needs review</div>
                 <div>{{sy-checkmark checked=activity.notBillable}} Not billable</div>
--- a/app/models/project.js
+++ b/app/models/project.js
@@ -51,6 +51,15 @@ export default Model.extend({
   customer: belongsTo("customer"),
 
   /**
+   * Whether the project's comments are visible to the customer
+   *
+   * @property customerVisible
+   * @type {Boolean}
+   * @public
+   */
+  customerVisible: attr("boolean", { defaultValue: false }),
+
+  /**
    * The billing
    *
    * @property {BillingType} billingType
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -291,3 +291,46 @@ strong {
     color: inherit;
   }
 }
+
+.activity-customer-visible-icon {
+  position: relative;
+  float: right;
+  z-index: 1;
+}
+
+.customer-visible-icon {
+  position: relative;
+  float: right;
+  margin-top: -1.7rem;
+  margin-right: 1rem;
+  z-index: 1;
+}
+
+@media (max-width: 680px) {
+  .activity-customer-visible-icon {
+    margin-top: 0rem;
+    margin-left : 90%;
+  }
+
+  .customer-visible-icon {
+    margin-top: -2.5rem;
+  }
+}
+
+@media #{$sm-viewport} {
+  .activity-customer-visible-icon {
+    margin-top: -2rem;
+    margin-left: 85%;
+  }
+}
+
+@media #{$md-viewport} {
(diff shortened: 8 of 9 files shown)


trowik:
Should I write some tests for each component to see if the icon is rendered?

trowik:
Random tests are failing each time

app/components/report-row/template.hbs
```diff
@@ -15,6 +15,11 @@
     onchange={{action (mut changeset.comment) value='target.value'}}
     data-test-report-comment
   >
+  {{#if changeset.task.project.customerVisible }}
+    <div class="customer-visible-icon">
```
aziiee, line 19: Could you please add a `title` to the `div`? Something like `This project is visible to the customer` to make it more explanatory.
And when you're at it, I would prefer if you could create a component for the implementation of the icon and reuse that everywhere else instead of creating a `div` each time manually. That way the code will be less redundant and more maintainable for the future.

app/models/project.js
```diff
@@ -50,6 +50,15 @@ export default Model.extend({
...
   customer: belongsTo("customer"),
 
+  /**
+   * Whether the project's comments are visible to the customer
+   *
+   * @property customerVisible
+   * @type {Boolean}
+   * @public
+   */
+  customerVisible: attr("boolean", { defaultValue: false }),
```
aziiee, line 60: Since it's a flag, I would prefer `isVisibleToCustomer` over `customerVisible`.

trowik: That doesn't work. It has to be `customerVisible`.

app/components/report-row/template.hbs
trowik: done

app/components/customer-visible-icon/component.js
```diff
@@ -0,0 +1,7 @@
+import Component from "@ember/component";
+
+export default Component.extend({
+  classNames: ["customer-visible-icon"],
+  attributeBindings: ["title"],
+  title: "This project's comments are visible to the customer"
+});
```
anehx, line 7: This needs to be a glimmer component.

app/components/customer-visible-icon/template.hbs
```diff
@@ -0,0 +1 @@
+{{fa-icon "eye"}}
```
anehx, line 1: Angle bracket syntax.. `<FaIcon @icon="eye" />`

app/components/report-row/template.hbs
```diff
@@ -15,6 +15,9 @@
     onchange={{action (mut changeset.comment) value='target.value'}}
     data-test-report-comment
   >
+  {{#if changeset.task.project.customerVisible }}
+      {{ customer-visible-icon }}
```
anehx, line 19: Angle bracket syntax

app/components/tracking-bar/template.hbs
```diff
@@ -24,6 +24,9 @@ as |t|}}
         value       = tracking.activity.comment
         disabled    = tracking.activity.active
       }}
+      {{#if tracking.activity.task.project.customerVisible }}
+          {{ customer-visible-icon }}
```
anehx, line 28: Dito

app/index/activities/edit/template.hbs
```diff
@@ -18,6 +18,9 @@
           value={{changeset.comment}}
           onchange={{action (mut changeset.comment) value='target.value'}}
         >
+        {{#if changeset.task.project.isVisibleToCustomer }}
+          {{ customer-visible-icon }}
```
anehx, line 22: Dito

app/index/activities/template.hbs
```diff
@@ -43,7 +43,12 @@
                   <strong>Unknown task</strong>
                 {{/if}}
               </td>
-              <td title={{activity.comment}}><div>{{activity.comment}}</div></td>
+              <td title={{activity.comment}}>
+                <div>{{activity.comment}}</div>
+                {{#if activity.task.project.customerVisible }}
+                  {{ customer-visible-icon }}
```
anehx, line 49: Dito

app/models/project.js
anehx, line 60: Refactor the whole model to class syntax if you touch it..

app/styles/app.scss
```diff
@@ -291,3 +291,47 @@ strong {
...
+    margin-left: 90%;
+  }
+}
+
+@media (max-width: 680px) {
+  .activity-customer-visible-icon {
+    margin-top: 0px;
+    margin-left: 85%;
+  }
+}
```
anehx, line 337: Those margins seem oddly specific, have you tried using `em` or `rem`?

anehx (changes requested):

app/components/report-row/template.hbs
anehx: `<CustomerVisibleIcon />` :wink:

app/models/project.js
aziiee: Yeah you're right, it depends on the [backend implementation](https://github.com/adfinis-sygroup/timed-backend/blob/master/timed/projects/models.py#L93)

app/components/customer-visible-icon/component.js
trowik: Did you mean something like that?
`import Component from "@glimmer/component";

export default class CustomerVisibleIcon extends Component {
  classNames = ["customer-visible-icon"];
  attributeBindings = ["title"];
  title = "This project's comments are visible to the customer";
}`
That didn't render the component and crashed others.
Besides that, why does it need to be a glimmer component? I couldn't find another component using glimmer.

app/components/customer-visible-icon/template.hbs
trowik: done

app/components/report-row/template.hbs
trowik: done

app/index/activities/edit/template.hbs
trowik: done

app/models/project.js
trowik: What do you mean exactly by that?
Something like that:

```
export defautl class Employment extends Model {

...

customer = belongsTo("customer");

customerVisible = attr("boolean");
}

app/styles/app.scss
trowik: No, I didn't. Changed it now.

app/index/activities/template.hbs
trowik: done

app/components/tracking-bar/template.hbs
trowik: done

aziiee:
@anehx We're still at ember v3.12 here which sounds like it's time for an upgrade. Maybe it would make more sense to first do the upgrade to the latest LTS in a separate PR and then gradually switch to octane? That way we could let this PR go through with the classic syntax. I'm not sure about the exact behavior in the app when switching to octane with ember < v3.15.

app/components/customer-visible-icon/component.js
anehx: Yeah, glimmer components don't render a DOM object. You need to create it in the template. See https://guides.emberjs.com/release/upgrading/current-edition/glimmer-components/#toc_outer-html for further infos

app/models/project.js
anehx: Not exactly: https://guides.emberjs.com/release/models/defining-models/#toc_defining-attributes

anehx:
> @anehx We're still at ember v3.12 here which sounds like it's time for an upgrade. Maybe it would make more sense to first do the upgrade to the latest LTS in a separate PR and then gradually switch to octane? That way we could let this PR go through with the classic syntax. I'm not sure about the exact behavior in the app when switching to octane with ember < v3.15.

Holy :hankey: Yes, an update to 3.24 is absolutely necessary. You can do it in another PR, but it needs to happen and it needs to happen now.

app/styles/app.scss
anehx: No idea how it looks, I didn't check it out :wink: However, CSS should be written a) mobile first and b) not with this many media queries for something as simple as margin. You can leave it, I don't really mind. Just try keeping it simple in the future.

app/components/customer-visible-icon/component.js
anehx: Since those features can't be used yet (https://github.com/adfinis-sygroup/timed-frontend/pull/562#issuecomment-867644369) you can leave it as it is. However, please create an issue to fix it after the update is done.

app/models/project.js
anehx: Same as https://github.com/adfinis-sygroup/timed-frontend/pull/562#discussion_r657992139

trowik:
Can this be merged?

anehx (approved):

anehx:
If you fix the tests yes

trowik:
That's the "problem" I have right now. That failing test is a random one, that only happens in CI and has nothing to do with the code changes. I don't have that problem locally and can't really reproduce it.
Any ideas on why that happens or maybe a possible solution?

PR #542, opened 2021-04-28
feat: add external employess

trowik:
add new boolean property to employment model
disable statistics page for external employees
disable analysis page for external employees
show contract type in user's general information box

PR #570, opened 2021-08-10
Refactor reviewers

trowik:
Use assignees instead of reviewers


trowik:
needs rebase after #542 is merged

PR #592, opened 2021-08-19
fix(projects): reset values when reapplying the filter

aziiee:
Project and task should be reset when the customer changes. Task
should be reset when the project changes.

### DIFF
--- a/app/projects/template.hbs
+++ b/app/projects/template.hbs
@@ -12,7 +12,7 @@
         options=customers
         placeholder="Select customer..."
         searchField="name"
-        onchange=(action (queue (action (mut selectedCustomer)) (action (mut selectedProject) null) (perform filterProjects)))
+        onchange=(action "handleCustomerChange")
         selected=selectedCustomer
         allowClear=true
       as |customer|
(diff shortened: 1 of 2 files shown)


app/projects/template.hbs
```diff
@@ -12,7 +12,7 @@
         options=customers
         placeholder="Select customer..."
         searchField="name"
-        onchange=(action (queue (action (mut selectedCustomer)) (action (mut selectedProject) null) (perform filterProjects)))
+        onchange=(action (queue (action (mut selectedCustomer)) (action (mut selectedProject) null) (action (mut selectedTask) null) (perform filterProjects)))
```
anehx, line 15: Please make an action in the controller for this. It's obviously not readable..

app/projects/template.hbs
```diff
@@ -26,7 +26,7 @@
         options=filteredProjects
         placeholder="Select project..."
         searchField="name"
-        onchange=(action (queue (action (mut selectedProject)) (perform fetchTasksOfProject)))
+        onchange=(action (queue (action (mut selectedProject)) (action (mut selectedTask) null) (perform fetchTasksOfProject)))
```
anehx, line 29: Same here

app/projects/template.hbs
aziiee: Done

app/projects/template.hbs
aziiee: Done

anehx (approved):

PR #596, opened 2021-08-26
fix(analysis): various optimizations

aziiee:
**Performance fix:**
Until now, all assignees were fetched which resulted into significant
performance loss on the analysis page. With this, we fetch assignees
by the given tasks, projects or customers of the current page only.

**Feature**
Display reports only if filters are set


velrest (approved):
Nice LGTM!

PR #604, opened 2021-09-09
feat: introduce the missing accountant role

aziiee:
An accountant is responsible for the billing of reports.
The feature "billing of reports" has already been implemented, but
we could not make use of it because of this missing role.

### DIFF
--- a/app/analysis/edit/template.hbs
+++ b/app/analysis/edit/template.hbs
@@ -4,7 +4,7 @@
     <div class="grid-cell">
       {{#if intersection.isRunning}}
         <div class="empty">
-          {{loading-icon}}
+          <LoadingIcon />
         </div>
       {{else}}
         <h1 class="text-center">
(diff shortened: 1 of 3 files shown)


app/analysis/edit/template.hbs
```diff
@@ -16,107 +16,173 @@
...
-                as |t|}}
-                  {{#f.input name='task' as |fi|}}
+            {{#let intersection.lastSuccessful.value.model as |model|}}
+              <ValidatedForm
+                @model={{changeset model IntersectionValidations}}
+                @on-submit={{perform save}}
+                as |f|
+              >
+                <TaskSelection
+                  @on-set-customer={{action (queue (action (mut f.model.customer)) (action 'validate' f.model))}}
```
velrest, line 26: Im not sure but is the first `action` helper here needed or could you just leave it out since out wrap all the calls in `action` anyway?

velrest:
Otherwise LGTM :ok_hand:

app/analysis/edit/template.hbs
aziiee: Good question.. but apparently it's needed: https://github.com/DockYard/ember-composable-helpers#queue

velrest: Okay then.

velrest (approved):

PR #608, opened 2021-09-15
chore: release v1.3.4

aziiee:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "timed",
-  "version": "1.3.2",
+  "version": "1.3.4",
```
aziiee, line 3: Jumping from `v1.3.2` to `v1.3.4` because the change in this file didn't happen at the time we released `v1.3.3`.

trowik (approved):
LGTM

PR #610, opened 2021-09-15
chore(deps): remove ember-cli-uglify

aziiee:
This dependency is not necessary and making trouble in production
builds.


trowik (approved):
Didn't we have issues with ember-cli-uglify last time as well?

aziiee:
> Didn't we have issues with ember-cli-uglify last time as well?

Yeah we replaced `ember-cli-uglify-es` at that time which is outdated for quite a while. I thought it was better to make use of the newer dependency, but since `uglify` is not really necessary and only causing troubles in our builds, I think it's best to get rid of it.

PR #624, opened 2021-10-06
fix(ability): can't edit reports if verified and reviewer

trowik:
You cannot update a report, if the corresponding project is set to billed.
I removed the billed permissions for now, as it cause us some problems.
Backend part for setting "billed"-flag isn't even implemented properly.


aziiee (approved):
LGTM

PR #637, opened 2021-11-02
feat(analysis): accountants may bill reports

trowik:
This will allow accountants to bill reports on analysis-edit. All
form elements except the `billed` checkbox will be disabled
if at least 1 report has been selected that was not created by the
accountant.

### DIFF
--- a/app/analysis/edit/controller.js
+++ b/app/analysis/edit/controller.js
@@ -65,6 +65,7 @@ export default Controller.extend(AnalysisEditQueryParams.Mixin, {
 
   analysisIndexController: controller("analysis.index"),
   intersectionModel: reads("intersection.lastSuccessful.value.model"),
+  isAccountant: reads("session.data.user.isAccountant"),
 
   setup() {
     this.get("intersection").perform();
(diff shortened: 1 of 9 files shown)


trowik:
Need to add tests

aziiee:
@trowik can you please also write an acceptance test in analysis-edit for this?

trowik:
Will do

aziiee:
Okay so with this, accountants will be able to edit everything in a report. How about restricting this so they can only edit the billed flag and everything else stays disabled? They should not be able to edit all other informations, that's the job of the assignees.

Can I also have your point of view on this @anehx @kaldras?

anehx:
> Okay so with this, accountants will be able to edit everything in a report. How about restricting this so they can only edit the billed flag and everything else stays disabled? They should not be able to edit all other informations, that's the job of the assignees.
> 
> Can I also have your point of view on this @anehx @kaldras?

@aziiee your approach seems more reasonable, let's do that.

trowik:
@aziiee ah yes. That's definitely a better approach

velrest:
@trowik whats the status with this pr?

trowik:
I thought that it was already merged, but it looks like it still needs a final review. @velrest

velrest (approved):

PR #651, opened 2021-12-21
feat: create view for users with no permission

aziiee:
Only users with an internal or external employment are allowed to
access timed. Users such as customers will be forwarded to the new
`no-access` page, where a message will be displayed that they do
not have the permission to access Timed.

### DIFF
--- a/app/protected/route.js
+++ b/app/protected/route.js
@@ -39,11 +39,15 @@ export default Route.extend(AuthenticatedRouteMixin, {
     const usermodel = await this.store.peekRecord("user", user.data.id);
 
     // Fetch current employment
-    await this.store.query("employment", {
+    const employment = await this.store.query("employment", {
       user: usermodel.id,
       date: moment().format("YYYY-MM-DD"),
       include: "location"
     });
+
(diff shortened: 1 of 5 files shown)


app/protected/route.js
```diff
@@ -39,12 +39,16 @@ export default Route.extend(AuthenticatedRouteMixin, {
...
 
     // Fetch current employment
-    await this.store.query("employment", {
+    const employment = await this.store.query("employment", {
       user: usermodel.id,
       date: moment().format("YYYY-MM-DD"),
       include: "location"
     });
 
+    if (!employment.length) {
```
trowik, line 48: In case of no employment `employment.length` is false?

trowik (approved):
Looks good to me. Did you test it?

app/protected/route.js
aziiee: yes

aziiee: I mean it's `0` in that case, which equals to false

aziiee:
> Looks good to me. Did you test it?

Sure did :)

PR #664, opened 2022-02-08
fix(auth): redirect to login on HTTP 401

aziiee:
Until now, a blank screen was shown due to an unhandled exception.
This should redirect to login if the request to get the current user
returns a 401.

### DIFF
--- a/app/services/ajax.js
+++ b/app/services/ajax.js
@@ -3,9 +3,9 @@
  * @submodule timed-services
  * @public
  */
-import { computed } from "@ember/object";
 import { reads } from "@ember/object/computed";
 import { inject as service } from "@ember/service";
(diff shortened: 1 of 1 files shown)


anehx:
1. Yes, I'm experiencing this bug all the time.
2. Your solution is partially correct. The 401 handling should happen in the ajax service - not where it's being used.
3. We should update at least the auth addons..

aziiee:
> 1. Yes, I'm experiencing this bug all the time.
> 2. Your solution is partially correct. The 401 handling should happen in the ajax service - not where it's being used.
> 3. We should update at least the auth addons..

@anehx The 401 is handled in the ajax service now. This definitely makes more sense.
There is a PR in progress for the octance upgrade. The auth addons will be updated there.

anehx (approved):

PR #709, opened 2022-10-21
Upgrade to ember v3.15

derrabauke:
## Contents

This bumps ember to v3.15 and `npx @ember/octanify` has been run.

---

PRs of this series will be merged into the `ember-octance` branch.
Later a PR for `ember-octance` onto `master` will follow.
[](https://mermaid.live/edit#pako:eNqdkjFvgzAQhf-KdV0TC8jGXFR1qBQRqV28OHCAFWwj9xhQlP9eg5vSQCpF9YS5995n6d4ZClsipFArenGya4Rh_hRWa0VMlakA61StjABGsvbXhMd8JyDojk6aomGoj-i2tiBp8DugweJke2Jaeu8ik4mJyjm_5tzMRguh8ck4zheBa5ZGV-MEYjR0mObZe5YfspsX7vNttAj69Wuih-8HabN5iYnXmHiNeQQQ333RPJiZyVqT_I-ZrPf_8bqf1_DEKmVk2w5_rjlE3WlEiAwl2vGIRwJgA17uvaVv4HnUCaAGtd_7WIQSK9m3NNIvXtp3pSTMSkXWQVrJ9hM3IHuyh8EUkJLr8Sp6VrJ2Uv-ocDK9hapPjb98AXwr8lw)


app/index/controller.js
```diff
@@ -144,7 +144,7 @@ export default Controller.extend({
         return total.add(moment().diff(current.get("from")));
       }, moment.duration());
 
-    this.set("_activeActivityDuration", duration);
+    // this.set("_activeActivityDuration", duration);
```
velrest, line 147: Is this safe to just comment out?

velrest, line 147: Is this safe to just comment out?

derrabauke: Well spotted! I will check!

derrabauke: >  ERROR: Error: Assertion Failed: You attempted to update `_activeActivityDuration` on `<timed@controller:index::ember9070>`, but it had already been used previously in the same computation.  Attempting to update a value after using it in a computation can cause logical errors, infinite revalidation bugs, and performance issues, and is not supported.

Ah :thinking: that's why he did it! 
The code in question:

```js
  /**
   * The duration sum of all activities of the selected day
   *
   * @property {moment.duration} activitySum
   * @public
   */
  activitySum: computed(
    "_activities.@each.{fromTime,duration}",
    "_activeActivityDuration",
    function() {
      this._activitySum();

      return this.get("_activities")
        .rejectBy("active")
        .reduce((total, current) => {
          return total.add(current.get("duration"));
        }, this.get("_activeActivityDuration"));
    }
  ),

  /**
   * Compute the current activity sum
   *
   * @method _activitySum
   * @private
   */
  _activitySum() {
    const duration = this.get("_activities")
      .filterBy("active")
      .reduce((total, current) => {
        return total.add(moment().diff(current.get("from")));
      }, moment.duration());

    this.set("_activeActivityDuration", duration);
  },
```

Commenting the line out fixes it indeed. But as soon as the class gets transformed into native class syntax + native `get`ters this line should be back in the code. I don't want to leave it like is, but postpone the transformation, since it will be included in future PR. Or do you see a better approach w/o refactoring to native classs+getters?

velrest (approved):

app/index/controller.js
velrest: No since its addressed anyway in the next few pr's i think its ok.

velrest: No since its addressed anyway in the next few prs i think its ok.

derrabauke: I tested it for a little, and deleting the `this._activitySum();` call, which is a side effect in the `activitySum` getter, fixed it. There is still some odd timing issues (wrong sum gets displayed for half a second when changing the focused task), but those will be fixed with the native class transformation. 

Could you please approve again @velrest :-)

PR #732, opened 2022-11-10
chore(deps): upgrade ember-power-calendar

derrabauke:
## Contents:
Upgrade ember-power-calendar and components which are based on it.
* 98ed572 refactor(sy-datepicker): octanify and fix tests
* 2c21088 refactor(date-navigation): octanify
* 30dfdbd chore(deps): resolve ember-concurrency version
* d14f7db chore(deps): upgrade eslint suite
* 8b9650e refactor(sy-calendar): refactor to octane
* 702e3ed chore(deps): ember-power-calendar

The contents of this PR start at 702e3ed
Commits before belong to #722

---


app/components/sy-datepicker/component.js
```diff
@@ -70,7 +73,18 @@ export default class SyDatepicker extends Component {
...
       const parsed = parse(value);
 
-      return this["on-change"](parsed && parsed.isValid() ? parsed : null);
+      return this.args.onChange(parsed && parsed.isValid() ? parsed : null);
     }
   }
+
+  @action
+  updateCenter({ moment }) {
+    return (this.center = moment);
```
velrest, line 82: Why is here a `return`?

velrest (commented):
Looking good otherwise

app/components/sy-datepicker/component.js
derrabauke: Don't know. Makes no sense.

velrest: If the return value is needed i would vote for 
```js
this.center = moment
return this.center
```
otherwise delete it.

derrabauke: `deleted`

velrest (approved):

PR #735, opened 2022-11-22
refactor:

derrabauke:
## Contents:
Refactor the custom ember-power-select implementation. This is an elementary part of the (activity) tracking view. 
* 1d0227e refactor(power-select): introduce optimized power select component
* ba3b073 refactor(tracking-view): refactor to octane

The contents of this PR start at ba3b073
Commits before belong to #732

---


derrabauke:
@velrest may I request your review for this wild `optimized-ember-power-select` stuff? :smiley: 

Here some hints to the eps architecture: https://ember-power-select.com/docs/architecture

app/components/optimized-power-select/options/component.js
```diff
@@ -0,0 +1,111 @@
+import { action } from "@ember/object";
+import Component from "@glimmer/component";
+
+const isTouchDevice = !!window && "ontouchstart" in window;
```
derrabauke, line 4: This seems to be copied form [the original eps options file](https://github.com/cibernox/ember-power-select/blob/64e91f272f2df87fe1cd5472abf596abfaa42ab1/addon/components/power-select/options.ts#L13)

Yelinz (approved):
Refactor looks good

PR #747, opened 2022-12-19
refactor(ember-fetch): use fetch instead of ajax

derrabauke:
## Contents
* 87b0210 refactor(ember-fetch): use fetch instead of ajax

---


app/services/fetch.js
```diff
@@ -0,0 +1,98 @@
...
+        ),
+      };
+    }
+
+    return await response.json();
+  }
+
+  get headers() {
+    const headers = {
+      accept: CONTENT_TYPE,
```
trowik, line 84: Not sure if missing quotation marks or shorthand?

app/services/fetch.js
```diff
@@ -0,0 +1,98 @@
...
+export default class FetchService extends Service {
+  @service session;
+
+  async fetch(resource, init = {}) {
+    init.headers = cleanObject({
+      ...this.headers,
+      ...(init.headers || {}),
+    });
+
+    if (init?.method === "POST" || init?.method === "PATCH") {
```
trowik, line 50: Shouldn't `PUT` method be stringified as well?
Or is it not used at all?

trowik (commented):
Some questions, but the rest looks good to me. (as far as I can tell)

trowik (approved):
Alright then. Looks good to me as far as I can tell.

PR #749, opened 2022-12-21
refactor: attendances, progress-tooltip and minor fixes

derrabauke:
## Contents

This has become a real *tutti frutti* PR with some smaller fixes and some refactorings.

* 4665c97 refactor(attendances): move actions to controller
* 2ffa8e2 refactor(attendance-slider): octanify and fix tests
* 3d4465c fix(sy-datepicker): prevent proxied date hicups
* 68e6c96 fix: temporary prevent error
* 8a37425 test: fix broken tests after codemods
* eda4288 fix: date/durationpicker usages in templates
* b64b62b fix(task-selection): reset selection correctly
* 034999c test: setup mirage for response mocking
* e688c25 refactor(progress-tooltip): octanify and fix tests
* 1180b9b refactor(sy-datepicker): improve UX and fix tests
* 20b197b test(report-row): refactor component invocation to octane
* b1338fd test(index-controller): remove unnecessary service stub
* 8eb6cd6 test(url-util): revert change in param parsing util

---


app/components/optimized-power-select/options/template.hbs
```diff
@@ -21,6 +21,7 @@
     @renderAll={{this.isTesting}}
     as |option index|
   >
+    {{!template-lint-disable  require-context-role}}
```
velrest, line 24: Why not set a `role`?

derrabauke:
```bash
 ember-template-lint --fix:
/home/falk/Dev/timed-frontend/app/components/optimized-power-select/options/template.hbs
  30:6  error  You have an element with the role of "option" but it is missing the required (immediate) parent element of "[listbox]". Reference: https://www.w3.org/TR/wai-aria-1.1/#option.  require-context-role
```
It actually has the role, but ember-template-lint is erroring incorrect, as it doesn't know about the `vertical-collection` wrapper.

app/index/attendances/controller.js
```diff
@@ -14,37 +15,102 @@ import AttendanceValidator from "timed/validations/attendance";
...
+   * Delete an attendance
+   *
+   * @method deleteAttendance
+   * @param {Attendance} attendance The attendance to delete
+   * @public
+   */
+  @action
+  async deleteAttendance(attendance) {
+    try {
+      await this.store.peekRecord("attendance", attendance.id).destroyRecord();
```
velrest, line 79: I think there is a `@service store` missing

mirage/factories/user.js
```diff
@@ -8,10 +8,24 @@ export default Factory.extend({
   email: () => faker.internet.email(),
   password: () => faker.internet.password(),
 
+  fullName() {
+    if (!this.firstName && !this.lastName) {
```
velrest, line 12: Maybe use ` !this.firstName || !this.lastName` otherwise you get `Jonas ` or ` Cosandey`

derrabauke: done

app/index/attendances/controller.js
derrabauke: done

velrest (approved):

PR #752, opened 2022-12-23
chore(lint): configure linting for gradual code adaption

derrabauke:
We want to trigger errors in `commitlint` and warnings in the `CI` for ember
deprecations, which do work for now (3.28 LTS) but will not work in the future
(>v4), to allow for gradual code adaption when touching the code.


Yelinz (approved):
nice work!

PR #760, opened 2022-12-27
chore: use pnpm in Dockerfile

trowik:
Correct me if I'm wrong, but `yarn` shouldn't be used as we switched to `pnpm`.
Not sure if I'm instaling it correctly though.

There's also an option to [fetch](https://pnpm.io/cli/fetch) packages to boost build time. Is this something we could use?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,16 @@
-FROM danlynn/ember-cli:latest as build
+FROM danlynn/ember-cli:3.28.0 as build
 
-COPY package.json yarn.lock /myapp/
+RUN npm install -g pnpm
 
-RUN yarn install
+COPY package.json pnpm-lock.yaml /myapp/
+
+RUN pnpm fetch
 
 COPY . /myapp/
 
-RUN yarn build --environment=production
+RUN pnpm install --frozen-lockfile --offline
+
+RUN pnpm run build --environment=production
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,12 +1,13 @@
 FROM danlynn/ember-cli:latest as build
 
-COPY package.json yarn.lock /myapp/
+COPY package.json pnpm-lock.yaml /myapp/
 
-RUN yarn install
+RUN npm install -g pnpm
```
anehx, line 5: Put this above the `COPY` command. Currently, this layer will be re-run if there are changes in `package.json` or `pnpm-lock.yaml`. However, this is not necessary because those files have no effect on this command.

Dockerfile
```diff
@@ -1,12 +1,13 @@
...
+RUN pnpm install
```
anehx, line 6: Use `--frozen-lockfile` to make sure there are no unintended changes in dependencies in the docker image.

Also, let's add `--prod` since we don't use the image for testing. This way we don't install the dev dependencies.

anehx (changes requested):
`pnpm fetch` looks like a great option to optimize the build time. Let's try it!

anehx (approved):

PR #762, opened 2023-01-02
ci: adapt test workflow to pnpm setup

derrabauke:
* add no-frozen-lockfile parameter to install step
* deletes workflow runs on ember-octane branch
* re-enables codecov reporting

PR #763, opened 2023-01-03
ci: upgrade workflow actions to use node 16 by default

derrabauke:
* Updates all actions to their latest version. This way we get `node 16` by default, still allowing for specific overrides.
* Further it improves the ci workflow by cleaning the `test` script defined in the `package.json`.
* Update and re-enable codecov action. This way future PRs have to at least improve the coverage and we are back in the coverage game. This leads to **failing CI tests for this PR!!** Please ignore the codecov jobs in this PR. The coverage compares with the last available upload which dates back to this commit d45179d.
* The formerly introduced change to the workflow triggers (`$default-branch` in #762 ) got reverted, since this variable is only available in workflow templates.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,17 +7,17 @@ jobs:
     name: Release
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


trowik (approved):
Great job!

PR #764, opened 2023-01-03
fix(filter-sidebar): raise z-index for overlay when visible

derrabauke:
## Issue
When using filters the screen isn't greyed out, leading to not being able to click anywhere on the screen to hide the filter menu.

## Solution
Raise z-index when overlay is visible.

### DIFF
--- a/app/styles/filter-sidebar.scss
+++ b/app/styles/filter-sidebar.scss
@@ -8,13 +8,13 @@
   min-width: 300px;
   max-width: 340px;
   flex-direction: column;
-  background: rgb(255,255,255);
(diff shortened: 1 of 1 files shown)


app/styles/filter-sidebar.scss
```diff
@@ -130,4 +132,5 @@
 .filter-sidebar-overlay--visible {
   pointer-events: all;
   opacity: 1;
+  z-index: 1;
```
derrabauke, line 135: This is the actual change. The rest is only **prettified**.

trowik (approved):
:+1:

PR #765, opened 2023-01-03
fix(analysis): table sorting according to ordering qp

derrabauke:
## Problem
Sorting the analysis table via clicking on the `Date` header does not trigger the ordering change.

## Solution
Do not serialize the `ordering` QP to `null`, if it's equal to the default value (`-date,id`)

## Background 
When entering `/analysis` the default query parameter for `ordering` should be `-date` which results in a serialized `-date,id` QP. As long as the QP is equal to it's default value, it should not occur in the URL. So far so good. 

But somehow, the QP will be present in the URL event though it's equal to the default value in the acceptance tests for `Acceptance | analysis edit: can edit`. :( 
Ergo we have to adapt the test-assertion or tune the QP more. What do you think @velrest ?

### DIFF
--- a/app/analysis/edit/controller.js
+++ b/app/analysis/edit/controller.js
@@ -180,7 +180,7 @@ export default class AnalysisEditController extends Controller.extend(
 
       this.router.transitionTo("analysis.index", {
         queryParams: {
-          ...params,
+          ...this.allQueryParams,
         },
       });
 
--- a/app/analysis/edit/controller.js
+++ b/app/analysis/edit/controller.js
@@ -86,14 +86,16 @@ export default class AnalysisEditController extends Controller.extend(
 
   @task
   *intersection() {
-    const res = yield this.fetch.fetch("/api/v1/reports/intersection", {
-      method: "GET",
-      data: {
+    const res = yield this.fetch.fetch(
+      `/api/v1/reports/intersection?${new URLSearchParams({
         ...prepareParams(this.allQueryParams),
         editable: 1,
         include: "task,project,customer,user",
-      },
-    });
+      })}`,
+      {
+        method: "GET",
+      }
+    );
 
     yield this.store.pushPayload("report-intersection", res);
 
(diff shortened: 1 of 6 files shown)


app/analysis/edit/controller.js
```diff
@@ -86,14 +86,16 @@ export default class AnalysisEditController extends Controller.extend(
 
   @task
   *intersection() {
-    const res = yield this.fetch.fetch("/api/v1/reports/intersection", {
-      method: "GET",
-      data: {
+    const res = yield this.fetch.fetch(
+      `/api/v1/reports/intersection?${new URLSearchParams({
```
derrabauke, line 90: Adding a body to a `GET` request is not spec conform. That's why we should append additional query parameters to the URL string instead.

app/analysis/index/controller.js
```diff
@@ -120,10 +120,10 @@ export const AnalysisQueryParams = new QueryParams({
     replace: true,
     refresh: true,
     serialize(val) {
-      return val === "-date" ? null : `${val},id`;
+      return val?.endsWith(",id") ? val : `${val},id`;
```
velrest, line 123: Wasn't the ` ? null` workaround to prevent the qp from resetting when editing reports?

derrabauke: yeah, that's at least what the comment states. But could not confirm this while testing with this new code. :shrug:

velrest: Huh, we spent so many hours trying to fix this and now it just works :sob:

velrest: But wont `val?.endsWith(",id") ` always be false since the `,id` part is only serialized?

derrabauke: > Huh, we spent so many hours trying to fix this and now it just **doens't** work sob

Sad! Indeed.

velrest (approved):

PR #766, opened 2023-01-03
fix(statistics): table sorting on header click

derrabauke:


app/statistics/template.hbs
```diff
@@ -214,7 +214,7 @@
         @type={{this.type}}
         @missingParams={{this.missingParams}}
         @ordering={{this.ordering}}
-        @on-ordering-change={{fn (mut this.ordering)}}
+        @onOrderingChange={{fn (mut this.ordering)}}
```
derrabauke, line 217: This is the actual fix of the "sorting on header click is not working".

app/components/sort-header/template.hbs
```diff
@@ -1,6 +1,9 @@
-{{yield}}
-{{#if this.active}}
-  <i class="fa fa-sort-{{this.direction}}"></i>
-{{else}}
-  <i class="fa fa-sort"></i>
-{{/if}}
+{{! template-lint-disable}}
```
velrest, line 1: What would template-lint criticize?

app/statistics/template.hbs
velrest: Sweet

velrest (approved):

app/components/sort-header/template.hbs
derrabauke: > app/components/sort-header/template.hbs
  1:24  warning  Interaction added to non-interactive element  no-invalid-interactive

PR #775, opened 2023-01-17
fix(auth): improve invalid session handling

derrabauke:
* redirect to `login` on `401` responses
* call correct action on `logout` click


trowik (approved):
LGTM

PR #787, opened 2023-01-27
fix: couple a bugs and improve UX

derrabauke:
**Fixes a couple of bugs after first review of pre-release by @czosel**
* 86588782 fix(sy-toggle): improve UI and keyboard navigation
* daed440c fix(fetch-service): handle invalid auth status
* 2062cd94 fix(tracking): initial fetch of drop down data
* b4332942 fix(statistic-list-bar): adapt css to different scenarios
* 8c315946 fix(statistics-list): calculate maxDuration in safe manner
* 2cbb3454 fix(report-row): make layout work for projects without rem eff

### DIFF
--- a/app/components/statistic-list/component.js
+++ b/app/components/statistic-list/component.js
@@ -65,7 +65,7 @@ export default class StatisticList extends Component {
   }
 
   get maxDuration() {
-    /* instanbul ignore if*/
+    /* istanbul ignore if*/
     if (!this.value) {
       return null;
     }
--- a/app/components/statistic-list/component.js
+++ b/app/components/statistic-list/component.js
@@ -71,10 +71,7 @@ export default class StatisticList extends Component {
     }
 
     const maxEstimated = moment.duration(
-      Math.max(
-        0,
-        ...this.value.filterBy("estimatedTime").mapBy("estimatedTime")
-      )
+      Math.max(0, ...this.value.mapBy("estimatedTime").filter(Boolean))
     );
     const maxDurationWithRemainingEffort = moment.duration(
       Math.max(
--- a/app/components/statistic-list/component.js
+++ b/app/components/statistic-list/component.js
@@ -71,7 +71,10 @@ export default class StatisticList extends Component {
     }
 
     const maxEstimated = moment.duration(
-      Math.max(0, ...this.value.mapBy("estimatedTime"))
+      Math.max(
+        0,
+        ...this.value.filterBy("estimatedTime").mapBy("estimatedTime")
+      )
     );
     const maxDurationWithRemainingEffort = moment.duration(
       Math.max(
--- a/app/components/sy-toggle/template.hbs
+++ b/app/components/sy-toggle/template.hbs
@@ -5,6 +5,11 @@
   tabindex="0"
   role="button"
   {{on "click" (optional (unless @disabled (fn (optional @onToggle))))}}
+  {{on-key
+    "Space"
+    (optional (unless @disabled (fn (optional @onToggle))))
+    onlyWhenFocused=true
+  }}
 >
   {{#if (has-block)}}
     {{yield}}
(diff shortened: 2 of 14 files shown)


velrest:
@derrabauke is this ready for review?

derrabauke:
> @derrabauke is this ready for review?

Now it is :)

app/components/statistic-list/component.js
```diff
@@ -71,7 +71,10 @@ export default class StatisticList extends Component {
     }
 
     const maxEstimated = moment.duration(
-      Math.max(0, ...this.value.mapBy("estimatedTime"))
+      Math.max(
+        0,
+        ...this.value.filterBy("estimatedTime").mapBy("estimatedTime")
```
velrest, line 76: Why are you calling `filterBy` without a value?

app/components/sy-toggle/template.hbs
```diff
@@ -5,6 +5,11 @@
   tabindex="0"
   role="button"
   {{on "click" (optional (unless @disabled (fn (optional @onToggle))))}}
+  {{on-key
+    "Space"
+    (optional (unless @disabled (fn (optional @onToggle))))
```
velrest, line 10: Why not just?
```suggestion
    @onToggle
```

velrest: To me it seems like calling the component `SyToggle` without the `@onToggle` doesn't make much sense. And if it does, there is an `optional` to much in that statement.

package.json
```diff
@@ -90,6 +90,7 @@
     "ember-event-helpers": "^0.1.1",
     "ember-export-application-global": "2.0.1",
     "ember-fetch": "8.1.2",
+    "ember-keyboard": "^8.2.0",
```
velrest, line 93: Not a fan of adding a whole dep for just one line `{{on "keyup"}}`  for example should also work

app/components/statistic-list/component.js
derrabauke: This filters out all `undefined` values. In case the list has only models without any entries with a `estimatedTime` attribute, it would return an empty array. `mapBy` on the other hand would return an array with `undefined` for each element.

```js
filterBy("estimatedTime") =>> []
mapBy("estimatedTime") =>> [undefined, undefined, ...] 
```

app/components/sy-toggle/template.hbs
derrabauke: The first one covers the case, that `@disabled` is `true` and the second `optional` is needed if `@onToggle` is undefined.
Normally one would use the constructor to assert `this.args.onToggle` is not empty, right?

package.json
derrabauke: My motivation was to keep it a *template only component*, but since the `this.args.onToggle` assertions would make the template also more readable, I guess it's time to create this `component.js` file. :sweat_smile:

app/components/statistic-list/component.js
velrest: Fair, as a nitpick if i get this correctly: `this.value.map(({estimatedTime})=>estimatedTime).filter(Boolean)` doesn't map the object by property twice. Or `this.value.mapBy("estimatedTime").filter(Boolean)` if you want to use `ArrayProxy` methods.

app/components/sy-toggle/template.hbs
velrest: > The first one covers the case, that @disabled is true and the second optional is needed if @onToggle is undefined.

I don't think `fn` is needed and then you only need an `optional` for the `{{on`.
> Normally one would use the constructor to assert this.args.onToggle is not empty, right?

Yep with `import { assert } from '@ember/debug';`

app/components/statistic-list/component.js
derrabauke: :thinking:

app/components/sy-toggle/template.hbs
derrabauke: Reworked. Plz check :)

velrest (approved):

PR #798, opened 2023-02-09
fix(customer-statistics): require fromDate param

derrabauke:
The customer-statistics request is quite demanding to for the backend. Especially on our test instance with only one pod this leads to very long waiting times or crashes of the backend. That's why we want to make the `fromDate` query param required for the customer statistics view.


trowik (approved):
LGTM

Issue #808, opened 2023-02-14
bug: refresh of absences not working properly


derrabauke:
> # Absences
> * After adding one or multiple absences they are not displayed until a refresh/switching day
> * After removing an absence, when trying to create a new absence (on the same day), the current day is selected and disabled, and therefore cannot be deselected.

Splitting it up for better issue tracking

_Originally posted by @derrabauke in https://github.com/adfinis/timed-frontend/issues/799#issuecomment-1429586584_

PR #809, opened 2023-02-14
refactor(sy-modal): glimmerify

derrabauke:
* refactoring of sy-modal component
* peripheral refactoring of welcome-modal and sy-modal-target components
* fix of #808


app/components/sy-modal/header/template.hbs
```diff
@@ -1,6 +1,11 @@
 <div ...attributes class="modal-header">
   {{yield}}
-  <button type="button" class="close" aria-label="Close" {{on "click" @close}}>
+  <button
+    type="button"
+    class="close"
+    aria-label="Close"
+    {{on "click" (optional @close)}}
```
luytena, line 7: Why is the close action optional?

app/components/sy-modal/overlay/template.hbs
```diff
@@ -1 +1,6 @@
-{{yield}}
+<div
+  class="modal modal-overlay {{if @visible 'modal--visible'}}"
+  {{did-insert this.setupClickHandler}}
```
luytena, line 3: Why can't we use something like `{{ on "click" this.handleClick }}`?

app/components/sy-modal/overlay/component.js
```diff
@@ -1,33 +1,23 @@
...
+    registerDestructor(this, () => {
+      element.removeEventListener("click", this.handleClick);
+    });
+  }
+
+  @action
+  handleClick(e) {
     if (e.target === this.element) {
-      this["on-close"]();
+      this.args.onClose();
```
luytena, line 20: Does this work if no `@onClose` is defined, since it's optional?

app/components/sy-modal/overlay/template.hbs
derrabauke: We need the components wrapping HTML-`element` to compare it in the on-click handler.

app/components/sy-modal/overlay/component.js
derrabauke: The optional helper replaces `undefined` with something like `() => {}`.

luytena: Oh I see :+1:

app/components/sy-modal/header/template.hbs
derrabauke: Artifact from previous refactoring attempt. Since the close action is wrapped before in a `optional` this should not be necessary anymore. Will delete.

derrabauke: `done`

luytena (approved):
LGTM :ok_hand:

PR #812, opened 2023-02-15
fix(remaining-effort): extract fresh info for tooltip

derrabauke:
## Description
The tooltip information was stale when task related remaining effort got updated after initial loading of the task/report page. 

## Implementation detail
I preferred to expand the `metadata-fetcher-service` for fetching fresh information on the project/task status instead of adding another request for updating the model in store. The meta-data request has to be made anyways and updating the model in question would only double up the request count. *Feedback welcome!*


Yelinz (approved):
Seems to be the only place where the method is called :+1:

PR #813, opened 2023-02-15
fix(reject-report): display rejected reports to responsible

derrabauke:
The frontend should warn the user who created the wrong report about the rejection. Not the reviewer.

Further this fixes the icon used for the warning.


trowik (approved):
LGTM

PR #814, opened 2023-02-15
fix(remaining-effort): display validation errors in project view

derrabauke:
We still need the `billingType` valdations for patches on the `projects` endpoint. To show the users they can't update the project (toggle remaining effort tracking checkbox), we show a warning icon with the error as hover description. See attached screenshot for final UI.

*warning icon on projects page*
error

*warning icon hovered*
error_title


trowik (approved):
Top :+1:

PR #822, opened 2023-02-16
fix(tracking-service): replace drop task modifier

derrabauke:
The `dropTask` modifier prevented multiple instances of the `TaskSelect` component in the timesheet overview from fetching their data for the dropdown. With the renewed logic, it will prevent fetching the same query multiple times.

### DIFF
--- a/app/services/tracking.js
+++ b/app/services/tracking.js
@@ -288,12 +288,21 @@ export default class TrackingService extends Service {
    * @param {Number} customer The customer id to filter by
    * @public
    */
-  @dropTask
+  @task
   *projects(customer) {
     if (!customer) {
       // We can't test this because the UI prevents it
       throw new Error("No customer selected");
     }
+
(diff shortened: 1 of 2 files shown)


app/services/tracking.js
```diff
@@ -305,13 +314,42 @@ export default class TrackingService extends Service {
...
+   * instance of it. Otherwise returns undefined.
+   */
+  filterDuplicateTasks(task, argument) {
+    const taskInstances = task.scheduler.taskInstances;
+    if (
+      taskInstances.length > 1 &&
+      taskInstances.filter((task) => task.args[0] === argument).length > 1
+    ) {
+      const original = taskInstances
+        .filter((task) => task.args[0] === argument)
```
czosel, line 348: just use `.find`

derrabauke:
@czosel Can we merge this as temporary fix and I will open the refactoring PR in an instant?

czosel (approved):
Of course :+1:

PR #827, opened 2023-02-20
fix(record-button): layout on different view port sizes

### LINKED ISSUE #820: bug: record-button and time incorrectly displayed
Current timed:
image
New timed:
image

derrabauke:
Fixes #820 

Includes:
* record button proportions on various screen sizes
* tidy up class definitions
* center play icon on record button via small transformation

### DIFF
--- a/app/components/record-button/template.hbs
+++ b/app/components/record-button/template.hbs
@@ -1,7 +1,4 @@
-<div
-  class="record-button-container record-button-container--start-stop
-    {{if this.active 'recording'}}"
->
+<div class="record-button-container {{if this.active 'recording'}}">
   {{#if this.active}}
     <button
       data-test-record-stop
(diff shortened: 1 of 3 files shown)


app/components/record-button/template.hbs
```diff
@@ -1,7 +1,4 @@
-<div
-  class="record-button-container record-button-container--start-stop
-    {{if this.active 'recording'}}"
->
+<div class="record-button-container {{if this.active 'recording'}}">
```
anehx, line 1: Let's use `record-button-container--recording` as active class since we mostly use BEM in our CSS: https://getbem.com/introduction/

anehx (changes requested):

app/components/record-button/template.hbs
derrabauke: Is it now `BEM`-like enough? :upside_down_face:

anehx (approved):

PR #828, opened 2023-02-20
chore(deps): patches and minor updates

derrabauke:


StephanH90 (approved):
As far as I can see only minor or patch versions have been bumped so should be ok I think.

PR #830, opened 2023-02-20
chore: bump displayed version to 3.0.0-rc

### LINKED ISSUE #825: Version displayed when using upstream is v2.1.3 which is not really true
When deploying the RC for 3.X the version displayed top left right below the `Timed` clock and name is still v2.1.3. But that should if a tag exists be the tag or at least be something like `upstream` or `dev` or whatever. Otherwise people are easily confused and might report bugs regarding a version which is not affecting this version and so on.

derrabauke:
Fixes #825 

`pre-releases` do not trigger semantic-release normally, so we have to bump it manually for now.

We'll investigate if we could leverage the `semantic-release` package also for the `pre-release` trigger.

See the [ember-cli-app-version docu](https://github.com/ember-cli/ember-cli-app-version#app-version-helper) for convenience.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "timed",
-  "version": "2.1.3",
+  "version": "3.0.0-rc.12",
   "private": true,
   "description": "A time tracking software written in Ember.js",
   "repository": "https://github.com/adfinis-sygroup/timed-frontend",
--- a/app/components/sy-topnav/template.hbs
+++ b/app/components/sy-topnav/template.hbs
@@ -12,7 +12,7 @@
       <div class="nav-top-header-title">
         Timed
         <div class="nav-top-header-title-version">
-          v{{app-version hideSha=true}}
+          v{{app-version versionOnly=true showExtended=true}}
         </div>
       </div>
     </LinkTo>


anehx:
Please add a number to the RC. There may be other release candidates for the 3.0.0 release.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "timed",
-  "version": "2.1.3",
+  "version": "3.0.0-rc12",
```
anehx, line 3: Pretty sure it's not the 12th release candidate :wink: Also, the rc version is delimited by a period.
```suggestion
  "version": "3.0.0-rc.1",
```

anehx (changes requested):

package.json
derrabauke: Well, we are hitting the twelfth pre-release... since this is the last tag [v3.0.0-rc.11](https://github.com/adfinis/timed-frontend/releases/tag/v3.0.0-rc11)

derrabauke: ... `rc` not used in an exemplary manner

anehx: Ah, I see. The package version was not being updated..

anehx (approved):

Issue #839, opened 2023-02-28
Most used tasks and projects


trowik:
Most used tasks and projects appear at the top of the selection and not in alphabetical order.

![image(1)](https://user-images.githubusercontent.com/55092761/221785491-2848a743-35ee-4345-b1e0-b9f8ccd430a6.png)

Could be connected to #838

derrabauke:
AFAIK this is intended behavior. Or I don't understand the report correctly.

trowik:
AFAIK the most used once should only appear in the customer selection and not in the other ones

derrabauke:
:eyeglasses: reading carefully. `noted` I will investigate into the issue!

derrabauke:
Fixed.

PR #842, opened 2023-02-28
fix(task-selection): respect archived filter

### LINKED ISSUE #838: Archived projects and tasks show up in selection
Archived tasks and project show up in selection for activities and timesheet.
This shouldn't be the case.
They should only show up in analysis and statistics filter.

image

derrabauke:
Fixes #838 and #839 

The previously refactored task-selection component, did not filter the respective projects and tasks for their current `archived` status. This yielded way more results in the dropdown as expect. Further those results where not ordered and that's why it seemed like it shows some *recent projects / tasks* which in reality, where just the first entries which got previously fetched to the store by other components of the app (e.g. timesheet).

### DIFF
--- a/app/components/task-selection/component.js
+++ b/app/components/task-selection/component.js
@@ -204,13 +204,11 @@ export default class TaskSelectionComponent extends Component {
 
     const customers = this.store
       .peekAll("customer")
-      .filter((customer) => {
-        return this.archived ? true : !customer.archived;
-      })
+      ?.filter(this.filterByArchived)
       .sortBy("name");
 
     const tasks = this.store.peekAll("task").filter((task) => {
-      return ids.includes(task.id) && (this.archived ? true : !task.archived);
+      return ids.includes(task.id) && this.filterByArchived(task);
     });
 
     return [...tasks.toArray(), ...customers.toArray()];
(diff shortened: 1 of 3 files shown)


app/components/task-selection/component.js
```diff
@@ -226,6 +226,22 @@ export default class TaskSelectionComponent extends Component {
...
+  }
+
+  get tasks() {
+    return this.project?.tasks
+      ?.filter((task) => {
+        return this.archived ? true : !task.archived;
+      })
+      .sortBy("name");
+  }
+
```
StephanH90, line 244:
```suggestion
  get projects() {
    return this.customer?.projects
      ?.filter(filterByArchived)
      .sortBy("name");
  }

  get tasks() {
    return this.project?.tasks
      ?.filter(filterByArchived)
      .sortBy("name");
  }

  filterByArchived(filterable) {
    return this.archived ? true : filterable.archived;
  }
```

if you wanted you could extract the duplicate filter function into a proper function instead.

StephanH90 (approved):
just 1 np otherwise LGTM

app/components/task-selection/component.js
derrabauke: Good points, thanks!

PR #850, opened 2023-03-01
fix(reports): include user to resolve name of verifier

### LINKED ISSUE #849: bug: tooltip on reviewed items says my boss is undefined
image

It's in the timesheet tab at `/reports`. I believe this to be a cosmetic bug, in analysis the same entry shows the right name.

derrabauke:
Fixes #849


trowik (approved):
:+1:

PR #854, opened 2023-03-02
fix(task-selection): fill selects with available entries

### LINKED ISSUE #852: bug: reporting doesn't show all projects of a customer
When trying to report something on Adfinis -> Marketing I can't see that project, I only see Classic IT and Talent Management.
Already tried to CTRL + F5 and opening timed in a private browser window, same problem.

2023-03-01-120027_1021x150_scrot

derrabauke:
Fixes #852 

When selecting a *recent task* (history) then we still need to load all available projects for the indirect selected customer. Otherwise the list of projects will be incomplete.


trowik (approved):
Pair reviewed

Yelinz (approved):

PR #856, opened 2023-03-02
fix: adding or editing tasks on projects page

### LINKED ISSUE #851: bug: tooltip remaining effort badge shows Infinity%
image

This project also doesn't have time tracking so there shouldn't be "remaining effort" at all.

The get request that is fired for the tooltip returns this:
```json
{
  "data": {
    "type": "tasks",
    "id": "3608727",
    "attributes": {
      "name": "teststask 123",
      "reference": "",
      "estimated-time": "00:00:00",
      "archived": false,
      "most-recent-remaining-effort": "05:00:00"
    },
    "relationships": {
      "project": {
        "data": {
          "type": "projects",
          "id": "4723"
        }
      },
      "cost-center": {
        "data": null
      }
    }
  },
  "meta": {
    "spent-time": "00:00:00"
  }
}
```

derrabauke:
Adding or editing tasks on projects page will no longer yield side effects, namely changing the `remaining-effort` of the task even if it was not enabled on the project.

Further fixes: 
* won't show remaining effort in progress tooltips
* repairs the table layout for the tasks of the project
* changes behavior of the `sy-durationpicker` component to allow values equal to `max` or `min`. That allows to `00:00` fields like the remaining effort input.

Fixes #851

### DIFF
--- a/app/components/progress-tooltip/component.js
+++ b/app/components/progress-tooltip/component.js
@@ -46,17 +46,22 @@ export default class ProgressTooltipComponent extends Component {
 
   get remainingEffort() {
     const model = this.args.model;
-    const isProjectWithEnabledTracking =
-      model.constructor.modelName === "project" &&
-      model.remainingEffortTracking;
+    const isProject = model.constructor.modelName === "project";
+    const hasRemainingEffortTracking = isProject
+      ? model.remainingEffortTracking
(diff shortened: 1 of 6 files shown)


Yelinz (approved):
Looks good

PR #869, opened 2023-03-14
fix: changeset was not trigger the form change (save button was not b…

### LINKED ISSUE #868: bug: activies edit mask won't show correct state for save button on activity change
image

Blue button should have correct state.

MitanOmar:
Fix #868


derrabauke (approved):
Looks good :+1:

PR #870, opened 2023-03-14
feat: gray background color for the archived projects

MitanOmar:
feature: the archived project will have gray background color instead of blue color


derrabauke (approved):
Good job!

PR #874, opened 2023-03-16
chore: upgrade ember-simple-auth-oidc

derrabauke:
https://github.com/adfinis/ember-simple-auth-oidc/releases


luytena (approved):
LGTM, but I have no experience with pnpm :)

Issue #819, opened 2023-02-16
bug: error messages on page load


c0rydoras:
On page refresh, sometimes i get the following:
Images:
![image](https://user-images.githubusercontent.com/110528300/219358139-a55fe698-d0ab-49e9-b367-df8a5fa5f30e.png)
![image](https://user-images.githubusercontent.com/110528300/219358589-a81edc69-f209-410a-9aac-045891c89da9.png)
![image](https://user-images.githubusercontent.com/110528300/219358475-ba83efad-1eb0-474f-baaa-89a111108933.png)
The errors:
```
Error while processing route: index.activities.index NetworkError when attempting to fetch resource. vendor-569ecb6afdc92a85e41011d096dc0804.js:2161:9
    error https://test.timed.adfinis.com/assets/vendor-569ecb6afdc92a85e41011d096dc0804.js:2161
    error https://test.timed.adfinis.com/assets/vendor-569ecb6afdc92a85e41011d096dc0804.js:2161
    z https://test.timed.adfinis.com/assets/vendor-569ecb6afdc92a85e41011d096dc0804.js:2171
    triggerEvent https://test.timed.adfinis.com/assets/vendor-569ecb6afdc92a85e41011d096dc0804.js:2095
    trigger https://test.timed.adfinis.com/assets/vendor-569ecb6afdc92a85e41011d096dc0804.js:4506
    transitionDidError https://test.timed.adfinis.com/assets/vendor-569ecb6afdc92a85e41011d096dc0804.js:2095
    promise https://test.timed.adfinis.com/assets/vendor-569ecb6afdc92a85e41011d096dc0804.js:4500
    v https://test.timed.adfinis.com/assets/vendor-569ecb6afdc92a85e41011d096dc0804.js:4683
    g https://test.timed.adfinis.com/assets/vendor-569ecb6afdc92a85e41011d096dc0804.js:4681
    h https://test.timed.adfinis.com/assets/vendor-569ecb6afdc92a85e41011d096dc0804.js:4679
    invoke https://test.timed.adfinis.com/assets/vendor-569ecb6afdc92a85e41011d096dc0804.js:4228
    flush https://test.timed.adfinis.com/assets/vendor-569ecb6afdc92a85e41011d096dc0804.js:4220
    flush https://test.timed.adfinis.com/assets/vendor-569ecb6afdc92a85e41011d096dc0804.js:4233
[...]

PR #879, opened 2023-03-22
chore(deps): upgrade ember-simple-auth-oidc to v5.0.3

### LINKED ISSUE #867: bug?: doing anything in timed when your session has expired doesn't redirect
# What happens
I'm not sure if this is intended behaviour, but its definitely different from before so I thought I would open a ticket anyway.

I never turn off my computer or browser so when I want to record my attendance when I start a new day I get an error message:

image

This is because my session has expired. When I reload the page, I instead get forward to the SSO sign in page. However, timed doesn't redirect me by itself, so to a non-techy user it might not be obvious that its actually the invalid session rather than some technical issue with timed.

# What I expect to happen
If my session is invalid/expired, and I try to do anything, I should be forwarded to a sign in screen.

derrabauke:
Fixes #867 #819

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "ember-route-action-helper": "2.0.8",
     "ember-scrollable": "1.0.2",
     "ember-simple-auth": "4.2.2",
-    "ember-simple-auth-oidc": "5.0.2",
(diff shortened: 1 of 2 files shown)


MitanOmar (approved):
nice :+1:

Issue #883, opened 2023-03-23
Attendances added to the wrong day after page refresh


C0rydoras:
## Issue
New attendance is not added correctly to the current day

https://user-images.githubusercontent.com/110528300/227140375-27f16c36-a1c5-4b7a-8719-8299325c3e74.mp4

Instead it is added to the current day
![image](https://user-images.githubusercontent.com/110528300/227140682-b910fe94-0687-4e2f-9bdf-dd6877c624d6.png)

## Steps to reproduce
* Go to the attendance Tab
* Choose a day other than the current day
* Refresh the page
* Now any new attendances will be added to the current day

PR #886, opened 2023-03-24
fix: attendances added wrong day after page refresh

MitanOmar:
fix: #883


derrabauke (approved):
This looks *LEAN*. :+1:

Issue #885, opened 2023-03-23
Negative Activities Possible


C0rydoras:
## Issue
Negative Activities are possible:
![image](https://user-images.githubusercontent.com/110528300/227143995-98648986-06c2-4970-accc-e2788e275e84.png)
![image](https://user-images.githubusercontent.com/110528300/227144614-795a2fee-dbbb-4a16-af91-60d50ed3ce16.png)

This completely breaks the `GENERATE TIMESHEET` feature, it stops rounding time values to be valid and also passes the negative Value to the Report.
![image](https://user-images.githubusercontent.com/110528300/227146758-6b6d162d-d45d-440c-b9d4-f1ebd021d944.png)

Since none of this is actually saved by the API  refreshing the page results in all invalid activities or reports being reverted or deleted.

These bugs appear to be relatively minor as they require very specific steps to reproduce and are not likely to occur during typical use of the application.
## How to reproduce
* Start an Activity
* Set the second input to a value lower than the first one.
* Click Save

Issue #881, opened 2023-03-22
Negative Attendance Possible


C0rydoras:
## Issue
It is possible to set negative attendance time
![image](https://user-images.githubusercontent.com/110528300/226893636-4023e82d-2d4b-4f62-80d8-580110e17d05.png)
## Steps to reproduce
- Go to the attendances tab
- Zoom in until the bar gets replaced by 2 Inputs
![image](https://user-images.githubusercontent.com/110528300/226895998-3c9f023c-b55a-400b-aef6-f984ac27fe26.png)
- These inputs are not validated for negative values. Therefore, negative values can be achieved by setting the second inputs value lower than the first inputs value.
![image](https://user-images.githubusercontent.com/110528300/226895577-e9bfde01-8521-444f-8b71-083a0ea88234.png)

MitanOmar:
fixed, and the fix will be available in the next release

PR #888, opened 2023-03-28
fix: negative activities possible

MitanOmar:
fix of #885 

the validation was checking the value in the changes, which was an empty object,
since an empty object was a valid value, (`undefined` and `null` are not acceptable), that's why the comparative was between moment instance and an empty object

### DIFF
--- a/app/validators/moment.js
+++ b/app/validators/moment.js
@@ -1,6 +1,13 @@
 import { get } from "@ember/object";
 import moment from "moment";
 
+function getDateTimeIfValid(momentObject) {
+  if (momentObject && momentObject._isValid) {
+    return momentObject;
+  }
+  return undefined;
+}
+
 /**
  * Validator to determine if a value is a valid moment object and if it is
  * greater or smaller than another property of the content.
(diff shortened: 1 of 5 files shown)


MitanOmar:
fix of #881 
since the fix is depend on some commits in this branch, the fix is pushed to this branch

app/validators/moment.js
```diff
@@ -19,23 +26,26 @@ export default function validateMoment(options = { gt: null, lt: null }) {
...
     if (options.gt) {
       const gtVal =
-        get(changes, options.gt) || get(content, options.gt) || moment();
-
+        getDateTimeIfValid(get(changes, options.gt)) ||
+        getDateTimeIfValid(get(content, options.gt)) ||
+        moment();
       if (newValue <= gtVal) {
-        valid = false;
+        valid = "false";
```
Yelinz, line 38:
```suggestion
        valid = `Value is smaller than ${options.gt}`;
```

app/validators/moment.js
```diff
@@ -19,23 +26,26 @@ export default function validateMoment(options = { gt: null, lt: null }) {
...
     if (options.lt) {
       const ltVal =
-        get(changes, options.lt) || get(content, options.lt) || moment();
+        getDateTimeIfValid(get(changes, options.lt)) ||
+        getDateTimeIfValid(get(content, options.lt)) ||
+        moment();
 
       if (newValue >= ltVal) {
-        valid = false;
+        valid = "false";
```
Yelinz, line 48:
```suggestion
        valid = `Value is larger than ${options.lt}`;
```

Yelinz (changes requested):

PR #894, opened 2023-04-03
fix: hbs warnings

MitanOmar:


app/users/edit/responsibilities/controller.js
```diff
@@ -7,11 +8,17 @@ import { all } from "rsvp";
...
+
   setup() {
     this.projects.perform();
     this.supervisees.perform();
   },
 
+  openSuperviseProfile: task(function* (superviseId) {
+    yield this.router.transitionTo("users.edit", superviseId);
+  }),
+
```
derrabauke, line 21: It should be supervis**or** I guess? ~~openSuperviseProfile~~ -> `openSupervisorProfile`.

Further, why using tasks instead of actions here?

app/users/index/controller.js
```diff
@@ -31,9 +31,14 @@ const UsersQueryParams = new QueryParams({
...
 const UsersIndexController = Controller.extend(UsersQueryParams.Mixin, {
   session: service("session"),
+  router: service("router"),
 
   currentUser: reads("session.data.user"),
 
+  viewUserProfile: task(function* (userId) {
+    yield this.router.transitionTo("users.edit", userId);
+  }),
+
```
derrabauke, line 41: ~~task~~ -> *action*

tests/acceptance/users-edit-responsibilities-test.js
```diff
@@ -9,7 +9,16 @@ module("Acceptance | users edit responsibilities", function (hooks) {
...
+    this.user = this.server.create("user", {
+      isSuperuser: true,
+      firstName: "Admin FirstName",
+      lastName: "Admin LastName",
+    });
+    this.secondUser = this.server.create("user", {
+      supervisorIds: [this.user.id],
+      firstName: "FirstName",
+      lastName: "LastName",
+    });
```
derrabauke, line 21: Nitpick: You could rename `this.user` --> `this.admin` and `this.secondUser` --> `this.user`. That way it would be more obvious what's the difference between those two users you've created. In this case the test is quite small and easy to grasp, but if it grows it becomes messy quite fast.

derrabauke (approved):
Good job! :tada:

PR #896, opened 2023-04-04
fix(sy-checkbox): add scss for indeterminate state

### LINKED ISSUE #871: [verifications] Indicator for entries that have the `needs review` flag set
My memory might be fuzzy, but I remember, when selecting multiple entries for verification, with some entries having the `needs review` flag set, there was an indication (`-`) on the `needs review` checkbox and I was able to remove the flag for all the entries.

Now the only way for me to know if `needs review` is set on some entries is the greyed out `verified` checkbox. As a workaround for removing the flag for all entries, I am able to set and then unset the `needs review` checkbox.

derrabauke:
Fixes #871

This was forgotten to be included in the refactoring process. :facepalm:

### DIFF
--- a/app/components/sy-checkbox/component.js
+++ b/app/components/sy-checkbox/component.js
@@ -1,8 +1,3 @@
-/**
- * @module timed
- * @submodule timed-components
- * @public
- */
 import { action } from "@ember/object";
 import { guidFor } from "@ember/object/internals";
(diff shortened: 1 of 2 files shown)


MitanOmar (commented):
nice

MitanOmar (approved):

PR #899, opened 2023-04-05
chore(dockerfile): pin pnpm to v7

derrabauke:
As long as we depend on `danlynn/ember-cli:3.28.5` we are bound to `node v16.13.1`. But since v8 `pnpm` requires at least `node v16.14`. As temporary resolution we have to pin `pnpm` to v7.

This is [blocking the ghcr image job](https://github.com/adfinis/timed-frontend/actions/runs/4617203705/jobs/8163238354).


trowik (approved):
:+1:

PR #900, opened 2023-04-11
fix: ember deprecation warnings

MitanOmar:


app/components/async-list/template.hbs
```diff
@@ -2,7 +2,7 @@
   <div class="empty">
     <LoadingIcon />
   </div>
-{{else if @data.last.isError}}
+{{else if (not @data)}}
```
derrabauke, line 5: `@data.isError` is a official [task state](https://ember-concurrency.com/api/TaskInstance.html) which is also available in here. It's also mentioned [here](https://ember-resources.pages.dev/funcs/util_ember_concurrency.trackedTask):

> (and all other properties on a [TaskInstance](https://ember-concurrency.com/api/TaskInstance.html))

tests/integration/components/async-list/component-test.js
```diff
@@ -47,7 +47,7 @@ module("Integration | Component | async list", function (hooks) {
   });
 
   test("shows error message", async function (assert) {
-    this.set("data", { last: { isError: true } });
+    this.set("data", false);
```
derrabauke, line 50: We have to adapt this to `isError` on the first level of the `data` object.

app/components/customer-visible-icon/component.js
```diff
@@ -1,9 +1,5 @@
-import Component from "@ember/component";
-import { attributeBindings } from "@ember-decorators/component";
-import classic from "ember-classic-decorator";
+import Component from "@glimmer/component";
 
-@classic
-@attributeBindings("title")
 export default class CustomerVisibleIcon extends Component {
   title = "This project's comments are visible to the customer";
```
derrabauke, line 4: Can't we make this a template only component by moving the title directly into the template?

app/services/tracking.js
```diff
@@ -135,10 +134,10 @@ export default class TrackingService extends Service {
...
       let task = "Unknown Task";
 
-      if (get(this, "activity.task.content")) {
-        const c = get(this, "activity.task.project.customer.name");
-        const p = get(this, "activity.task.project.name");
-        const t = get(this, "activity.task.name");
+      if (this.activity.task.content) {
+        const c = this.activity.task.get("project.customer.name");
+        const p = this.activity.get("task.project.name");
+        const t = this.activity.get("task.name");
```
derrabauke, line 140: This seems a little inconsistent. Why don't we need the get for the first `task` access?

MitanOmar: it was throwing errors without it

app/users/index/controller.js
```diff
@@ -1,106 +1,101 @@
...
-
-  reset() {
-    /* istanbul ignore next */
-    this.resetQueryParams();
-  },
+  }
+
+  _fetchData = trackedTask(this, this.data, () => {
+    return [this.supervisor, this.search, this.ordering, this.active];
+  });
```
derrabauke, line 27:
```suggestion
  _fetchData = trackedTask(this, this.data, () => [this.supervisor, this.search, this.ordering, this.active]);
```

app/users/index/controller.js
```diff
@@ -1,106 +1,101 @@
...
-    }
-  ),
-
-  setup() {
+export default class UsersIndexController extends Controller {
+  queryParams = ["search", "supervisor", "active", "ordering"];
+  @service session;
+  @service router;
+  @service store;
+  @tracked currentUser = this.session.data.user;
```
derrabauke, line 14: This is not the same as `currentUser: reads("session.data.user")`

```suggestion
  get currentUser  () {
    return this.session.data.user
  }
```

app/users/index/controller.js
```diff
@@ -1,106 +1,101 @@
...
-    defaultValue: "1",
-    replace: true,
-    refresh: true,
-  },
-  ordering: {
-    defaultValue: "username",
-    replace: true,
-    refresh: true,
-  },
-});
```
derrabauke, line 30: The previous `ember-parachute` implementation had the `refresh` and `replace` parameters set to true. Did you have a look what they do? 

I guess we have to add the [`refreshModel` parameter](https://guides.emberjs.com/release/routing/query-params/#toc_opting-into-a-full-transition) on the refactored route.

app/users/index/controller.js
```diff
@@ -1,106 +1,101 @@
...
-      ...(this.get("currentUser.isSuperuser")
+      ...(this.currentUser.isSuperuser
         ? {}
         : {
-            supervisor: this.get("currentUser.id"),
+            supervisor: this.currentUser.id,
           }),
     });
-  }).restartable(),
+  }
```
derrabauke, line 88: We could implement this with the route (model) and the native query params, too. Or use 
 [ember-async-data](https://github.com/tracked-tools/ember-async-data). This whole controller feels so overcomplicated. :man_shrugging:

derrabauke (approved):
LGTM :eyeglasses:

PR #903, opened 2023-04-13
fix(attendance-slider): update validation status before saving

### LINKED ISSUE #902: bug: attendance slider jumps back to initial position after drag
## Description
Dragging a attendance slider to a later time will not suffice. It will jump back to it's initial position. This only happens if the new **start-time** of the slider is beyond the **latest end-time** of the slider.
image

## Workaround
Increase the overall time (by dragging the right-hand side handle of the slider) of your block to the intended end time. Adapt the start time accordingly.

derrabauke:
Fixes #902


MitanOmar (approved):

mjoxy:
probably also solves #904

PR #909, opened 2023-04-19
fix(protected): await before model hook auth request

derrabauke:
We should await the `auth` response, before proceeding with further requests in the protected route. Otherwise it is more prune for race-conditions in the auth flow. Full description see: https://github.com/adfinis/ember-simple-auth-oidc/pull/761

### DIFF
--- a/app/protected/route.js
+++ b/app/protected/route.js
@@ -24,8 +24,8 @@ export default class ProtectedRoute extends Route {
   @service store;
   @service fetch;
 
-  beforeModel(transition) {
-    this.session.requireAuthentication(transition, "login");
+  async beforeModel(transition) {
+    await this.session.requireAuthentication(transition, "login");
   }
 
   async model() {


anehx (approved):

czosel:
Is this method really async? https://github.com/mainmatter/ember-simple-auth/blob/0b4bebb95a74080cc20256ebd65cbb7e5890611f/packages/ember-simple-auth/addon/services/session.js#L229

czosel:
Also, what exactly is the `auth` response? Do you mean the `/me` endpoint? The handshake with the SSO happens via redirects, not AJAX calls.

derrabauke:
> Is this method really async? https://github.com/mainmatter/ember-simple-auth/blob/0b4bebb95a74080cc20256ebd65cbb7e5890611f/packages/ember-simple-auth/addon/services/session.js#L229

You were locking at the wrong location. [Have a look at this](https://github.com/adfinis/ember-simple-auth-oidc/blob/b798d8578b7e3523606d92d69575bea8fd9272df/addon/services/session.js#L66).

PR #910, opened 2023-04-20
chore: upgrade ember-simple-auth-oidc

### LINKED ISSUE #907: bug: SSO token handling
The SSO token handling still feels degraded compared to before the last big update end of February. The frontend stays active but is no longer able to access the backend du to an expired token. This sometimes leads to the situation that changes can't be written and are lost after a refresh of the page that redirects to the oidc auth. Other pages of course also misbehave because of this behavior.

At least, I'd expect the frontend to redirect to oidc auth as soon as the token's expired. Even cooler would be to cache unsaved changes somehow and keep them until re-authentication. Don't know if that is possible.

derrabauke:
Fixes #907

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
     "ember-route-action-helper": "2.0.8",
     "ember-scrollable": "1.0.2",
     "ember-simple-auth": "4.2.2",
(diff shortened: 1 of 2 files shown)


trowik (approved):
:tada:

PR #915, opened 2023-04-24
feat: login page

MitanOmar:
create a login page, and move the `sso-login` to new route `/sso-login`

## Design Preview
image


app/login/controller.js
```diff
@@ -0,0 +1,12 @@
+import Controller from "@ember/controller";
```
derrabauke, line 1: Not needed.

app/login/template.hbs
```diff
@@ -0,0 +1,11 @@
+<div class="login">
+  <h1>Timed Login</h1>
+  <hr>
+  <TimedClock />
+  <br>
+  <button
+  class="btn btn-primary sso-login"
+  type="button"
+  {{on "click" this.ssoLogin}}
+  >Login With SSO</button>
```
derrabauke, line 10: Use [LinkTo](https://guides.emberjs.com/release/routing/linking-between-routes/) component with button styling instead.

derrabauke:
Could you please edit the last commit message 

from:
* ~~[fix: use linkedTo instate of routing via controller](https://github.com/adfinis/timed-frontend/pull/915/commits/2d060d9b1513ddc0843fe177c32f3e1a2a7990c8)~~

to: 
* fix: use link-to component for routing

**or**
amend the second commit to the first one.

derrabauke:
We should add `afterLogoutUri: "/sso-login",` to the `environment.js` config.

app/styles/login.scss
```diff
@@ -0,0 +1,3 @@
+.login {
+  text-align: -webkit-center;
```
derrabauke, line 2: Is the `-webkit` prefix really needed?

derrabauke:
Maybe add a another acceptance test, where we check if the click on the logout button is really redirecting to the *login* page. :sunglasses:

app/styles/login.scss
MitanOmar: it's getting all elements to be centered, other options all not centering everything

derrabauke: It know what it does, but we don't need the `-webkit` [vendor prefix](https://developer.mozilla.org/en-US/docs/Glossary/Vendor_Prefix). This is a standard css property which is implemented the same across the latest major browsers.

derrabauke:
@MitanOmar @kaldras: I had a bilateral discussion with @czosel about the login page design. After that we directly implemented the changes.

* increase margin for title
* increase clock size
* make headline a little more friendly `Welcome to Timed`

What do you guys think about it? Thumbs up if we can merge it that way. :-)

MitanOmar:
nothing to add from my side, it look great

derrabauke:
Sorry, for the *push-force* noise. Rebased on `main`, but target should be `beta`, so had to do it again. :man_shrugging:

derrabauke (approved):

kaldras (approved):

PR #917, opened 2023-04-25
Fix js warning

MitanOmar:
use native JS classes in ember models


MitanOmar:
this PR is moved to #921

Issue #918, opened 2023-04-25
feature: improve statistics bar chart rendering


derrabauke:
Projects and tasks with a defined remaining effort value, will render the goal line and visually indicate if the budget got overspent (marking the bar red). Projects/Tasks without a given *remaining effort* won't receive another color, even when spent over budget. To align the behavior for both scenarios we want to introduce some light and dark color shades.

**Dark color shades:**
* blue: already spent efforts in budget
* green: [finished/archived] in budget
* red: [already spent efforts/finished/archived] over budget

**Light color shades:**
* green: remaining effort (+ already spent efforts in blue) **in range** of expected budget
* red: remaining effort (+ already spent efforts in blue) **exceeding** the expected budget

**A mockup of the different scenarios:**
![image](https://user-images.githubusercontent.com/10029904/234274951-eb1a8c2a-ea1a-4478-9f23-b6ff1ff99507.png)

PR #923, opened 2023-04-26
feat: improve statistics bar chart rendering

MitanOmar:


MitanOmar:
This PR includes feature request of #918

app/components/statistic-list/bar/component.js
```diff
@@ -0,0 +1,17 @@
+import Component from "@glimmer/component";
+
+export default class StatisticListBar extends Component {
+  get didFinishEffortsInBudget() {
+    if (this.args.remaining === 0 && this.args.value < this.args.goal) {
+      return true;
+    }
+    return false;
+  }
```
derrabauke, line 9: We need a special flag `archived` on the task / project to realize this feature. This has to be done in the backend. Otherwise all running tasks, which don't have remaining effort defined are colored :green_circle:, which is not intended.

Let's skip the :green_circle: state for now. We have to discuss the efforts for this first.

tests/integration/components/statistic-list/bar/component-test.js
```diff
@@ -18,4 +18,30 @@ module("Integration | Component | statistic list/bar", function (hooks) {
       "0.5"
     );
   });
+
+  test("The Chart color is RED when spent effort is over the budget", async function (assert) {
```
derrabauke, line 22: We don't have to *yell* the color. ;)

tests/integration/components/statistic-list/bar/component-test.js
```diff
@@ -18,4 +18,30 @@ module("Integration | Component | statistic list/bar", function (hooks) {
...
+    await render(
+      hbs`<StatisticList::Bar @value={{0.5}} @goal={{0.4}} @remaining={{0}} />`
+    );
+
+    const element = this.element.querySelector(".statistic-list-bar");
+
+    assert.ok(element);
+
+    assert.true(element.classList.contains("strong-danger"));
+    assert.false(element.classList.contains("strong-success"));
```
derrabauke, line 32:
```suggestion
    assert.true(element.classList.contains("strong-danger"));
```

app/components/statistic-list/bar/component.js
```diff
@@ -0,0 +1,17 @@
...
+
+  get didFinishEffortsOverBudget() {
+    if (this.args.value > this.args.goal) {
+      return true;
+    }
+    return false;
+  }
```
derrabauke, line 16:
```suggestion
  get didFinishEffortsOverBudget() {
    return this.args.value > this.args.goal;
  }
```

app/components/statistic-list/bar/component.js
derrabauke, line 9:
```suggestion
  get didFinishEffortsInBudget() {
    return this.args.archived && !this.didFinishEffortsOverBudget;
  }
```

tests/integration/components/statistic-list/bar/component-test.js
derrabauke, line 1: Let's add test for all the different states as well!

| Spent |  Budget |  RE |  Effort Bar | RE Bar  |
|---|---|---|---|---|
| 1  | 1  |  0 |  :blue_square:  | - |
| 2  | 1  |  0 | :red_square:   | -  |
|  ¼ | 1  |  ½ | :blue_square:   | :green_circle:  |
| ½ | 1 | 1 | :blue_square:  | :red_square: |
|1 | ½ | 1 | :red_square:  | :red_square: |

MitanOmar: those cases also are covered in the tests

app/components/statistic-list/bar/component.js
```diff
@@ -0,0 +1,21 @@
...
+
+  get didFinishEffortsOverBudget() {
+    return this.args.value > this.args.goal;
+  }
+
+  get spentEffortsBarColor() {
+    if (this.didFinishEffortsInBudget) {
+      return "strong-success";
+    }
```
derrabauke, line 15: This is still coloring the bar :green_circle: if the currently spent effort is less than the estimated goal and if not remaining effort is set. But we discussed, this should only get green, if the archived flag is `true`. As we have no archived flag yet, *no report should be green*. Only blue like in the old days. I suggest you already add the check for the `archived` flag. As it's not given yet by the backend it should always be `false` anyways. So we won't have to change anything if it gets enabled in the future. What you think?

derrabauke: image

app/components/statistic-list/bar/component.js
```diff
@@ -0,0 +1,21 @@
...
+  get didFinishEffortsInBudget() {
+    return this.args.remaining === 0 && !this.didFinishEffortsOverBudget;
+  }
+
+  get didFinishEffortsOverBudget() {
+    return this.args.value > this.args.goal;
+  }
+
+  get spentEffortsBarColor() {
+    if (this.didFinishEffortsInBudget && this.args.archived) {
```
derrabauke, line 13:
```suggestion
  get didFinishEffortsInBudget() {
    return this.args.remaining === 0 && !this.didFinishEffortsOverBudget && this.args.archived;
  }

  get didFinishEffortsOverBudget() {
    return this.args.value > this.args.goal;
  }

  get spentEffortsBarColor() {
    if (this.didFinishEffortsInBudget) {
```

derrabauke (approved):
:+1: good job

MitanOmar:
finally :)

PR #931, opened 2023-04-27
refactor: not identical warning to glimmer component

MitanOmar:


derrabauke (changes requested):
Please update component test file to angle brackets.

Issue #939, opened 2023-05-03
Verified Reports are Editable


C0rydoras:
# Issue
Validated Reports are editable.
The problem seems to be with the [abilities](https://github.com/adfinis/timed-frontend/blob/c98e9aa8bbbc92724d79eb14f41fedebf00bc88a/app/abilities/report.js) as this here:
```js
  get editable() {
    return this.abilities.can("edit report", this.args.report);
  }
```
returns true.

C0rydoras:
@MitanOmar

MitanOmar:
@C0rydoras i will back to it ASAP

PR #942, opened 2023-05-10
fix: verified reports shold not be editable

MitanOmar:
fix of #939

### DIFF
--- a/app/abilities/report.js
+++ b/app/abilities/report.js
@@ -11,13 +11,13 @@ export default Ability.extend({
     function () {
       const isEditable =
         this.user?.isSuperuser ||
-        (!this.model.verifiedBy?.id &&
+        (!this.model?.verifiedBy?.get("id") &&
           // eslint-disable-next-line ember/no-get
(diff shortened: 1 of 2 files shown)


app/abilities/report.js
```diff
@@ -11,21 +11,21 @@ export default Ability.extend({
...
+              .includes(this.user?.get("id"))));
       const isReviewer =
         (this.model?.taskAssignees ?? [])
           .concat(
             this.model?.projectAssignees ?? [],
             this.model?.customerAssignees ?? []
           )
           .mapBy("user.id")
-          .includes(this.user?.id) && !this.model.verifiedBy?.id;
+          .includes(this.get("user.id")) && !this.model?.get("verifiedBy.id");
```
derrabauke, line 28:
```suggestion
          .includes(this.user.get("id")) && !this.model?.get("verifiedBy.id");
```
Why is the access of `user.id` here different from the one above?

derrabauke (approved):

PR #949, opened 2023-05-24
refactor(balance-donut): to glimmer sytax

MitanOmar:


app/components/balance-donut/component.js
```diff
@@ -1,32 +1,21 @@
...
 const { PI, floor, min, abs } = Math;
 
 const { isInteger } = Number;
-
-@classic
-@attributeBindings("style")
-@classNameBindings("color")
 class BalanceDonutComponent extends Component {
-  @computed("balance.{usedDays,usedDuration,credit}")
+  balance = this.args.balance;
```
derrabauke, line 8: What's that for? Assigning this once would thwart reactivity.

derrabauke: If this is for shortening property access then use a getter.

app/components/balance-donut/component.js
```diff
@@ -43,13 +32,12 @@ class BalanceDonutComponent extends Component {
 
   radius = 100 / (2 * PI);
 
-  @computed("count", "index")
   get style() {
```
derrabauke, line 35: This is a point where we could consider using [the cached decorator](https://api.emberjs.com/ember/release/functions/@glimmer%2Ftracking/cached), since we do a lot of computation in this getter.

app/components/balance-donut/template.hbs
```diff
@@ -1,40 +1,42 @@
-<div class="donut-title">{{this.title}}</div>
+<div style={{this.style}} class="user-header-absence-balance balance-donut {{this.color}}">
```
derrabauke, line 1:
```suggestion
<div {{style this.style}} class="user-header-absence-balance balance-donut {{this.color}}">
```

Let's use the [ember-style-modifier](https://github.com/jelhan/ember-style-modifier) since we already have it.

derrabauke, line 1:
```suggestion
<div style={{this.style}} class="balance-donut {{this.color}}">
```

What is this `user-header-absence-balance` class for? Seems like a use case specific class which should not be in here.

app/styles/app.scss
```diff
@@ -47,6 +47,7 @@
 @import "components/filter-sidebar--label";
 @import "components/filter-sidebar--group";
 @import "components/date-buttons";
+@import "components/balance-donut.scss";
```
derrabauke, line 50:
```suggestion
@import "components/balance-donut";
```

derrabauke (changes requested):

derrabauke (approved):

PR #956, opened 2023-05-31
fix: weekly overview benchmark tests

MitanOmar:


tests/integration/components/weekly-overview-benchmark/component-test.js
```diff
@@ -15,10 +15,7 @@ module("Integration | Component | weekly overview benchmark", function (hooks) {
   test("computes the position correctly", async function (assert) {
     await render(hbs`{{weekly-overview-benchmark hours=10 max=10}}`);
 
-    assert.strictEqual(
-      find("hr").getAttribute("style"),
-      "bottom: calc(100% / 10 * 10)"
-    );
+    assert.strictEqual(find("hr").getAttribute("style"), "bottom: calc(100%);");
```
derrabauke, line 18: I don't get it. Why is bottom value `calc(100%)`?

MitanOmar: i think because [ember-style-modifier](https://github.com/jelhan/ember-style-modifier) is calculating it before sending it to the template

derrabauke: [It does indeed](https://github.com/jelhan/ember-style-modifier/blob/73d9a1c324bda11991cda21785a5b062abfd6456/addon/modifiers/style.js#LL57C11-L58C1)! 

The key point here is how it sets the property on the HTML element: `element.style.setProperty("bottom", value)` *vs* `element.setAttribute("bottom", value)

derrabauke (approved):
:+1:

PR #965, opened 2023-06-06
refactor(activity-model): use modern class syntax

MitanOmar:


app/models/activity.js
```diff
@@ -72,15 +66,15 @@ export default Model.extend({
...
+  @belongsTo("user") user;
 
   /**
    * The notify service
    *
    * @property {EmberNotify.NotifyService} notify
    * @public
    */
-  notify: service("notify"),
+  @service("notify") notify;
```
derrabauke, line 77:
```suggestion
  @service notify;
```

derrabauke (changes requested):
Just some nitpicks:
* proposal for a better commit message `refactor(activity-model): use modern class syntax`
* delete outdated doc string

derrabauke (approved):

PR #967, opened 2023-06-08
refactor(progress-bar): component to use glimmer syntax

MitanOmar:


app/components/progress-bar/component.js
```diff
@@ -39,6 +28,11 @@ class ProgressBarComponent extends Component {
    */
   color = null;
 
+  constructor(...args) {
+    super(...args);
+    this.progress = this.args.progress || 0;
+  }
+
```
derrabauke, line 35: Why do you want to clone the state (`progress`) in the component? Just pass it with a getter for example:
```js
get progress() {
  return this.args.progress || 0;
}
```

derrabauke (changes requested):

derrabauke (approved):

PR #994, opened 2023-06-15
Replace ember site tour

MitanOmar:

PR #996, opened 2023-06-21
refactor(activites-route): to use native js classes

MitanOmar:


MitanOmar:
this can not be merged before refactoring the tours in #994

Issue #786, opened 2023-01-26
Refactor deprecated code


derrabauke:
The following list of deprecations must be processed.
 
`deprecation-workflow.js`:
```
    { handler: "silence", matchId: "ember-global" },
    { handler: "silence", matchId: "ember.component.reopen" },
    { handler: "silence", matchId: "ember-metal.get-with-default" },
    { handler: "silence", matchId: "implicit-injections" },
    { handler: "silence", matchId: "ember-views.curly-components.jquery-element" },
    { handler: "silence", matchId: "deprecated-run-loop-and-computed-dot-access" },
    { handler: "silence", matchId: "ember-utils.try-invoke" },
    { handler: "silence", matchId: "this-property-fallback" },
    { handler: "silence", matchId: "autotracking.mutation-after-consumption" },
    { handler: "silence", matchId: "manager-capabilities.modifiers-3-13" },
    { handler: "silence", matchId: "setting-on-hash" },
    { handler: "silence", matchId: "routing.transition-methods" },
```
## How to:
Run `ember test -s` and delete a line from `deprecation-workflow.js` and fix the error's that occur. When finished with one delete the next line and so on. 

--- 
## PRs
#785

MitanOmar:
Fixed with PR #893

derrabauke:
I think we are not done yet. ;)

PR #997, opened 2023-06-21
fix: implication injections in deprecation workflow

MitanOmar:
this is the final part of #786


app/abilities/report.js
```diff
@@ -1,32 +1,34 @@
...
-        this.user?.isSuperuser ||
-        (!this.model?.verifiedBy?.get("id") &&
+export default class ReportAbility extends Ability {
+  @service session;
+  @tracked user;
+
+  constructor(...args) {
+    super(...args);
+    this.user = this.session.data.user;
+  }
```
derrabauke, line 12: Why not using a getter for the `user` as in the other ability classes?

config/deprecation-workflow.js
```diff
@@ -3,7 +3,7 @@ window.deprecationWorkflow = window.deprecationWorkflow || {};
 window.deprecationWorkflow.config = {
   workflow: [
     { handler: "silence", matchId: "ember.component.reopen" }, // not finish yet
-    { handler: "silence", matchId: "implicit-injections" },
+    { handler: "throw", matchId: "implicit-injections" },
```
derrabauke, line 6: We can delete this line after all errors are resolved. :relaxed:  :+1:

app/abilities/report.js
MitanOmar: because of tests in `tests/unit/abilities/report-test.js` file
this file is setting the value of `user` property

derrabauke: I think the test should not be the priority for a the actual implementation. Let's try it with a `get user ()` in the ability and change the test to use `setupSessionMock`

```js
import setupSession from "timed/tests/helpers/session-mock";

// in the test module
setupSession(hooks)

// in the test case
const session = this.owner.lookup("service:session");
session.data.user = { ...whateverisneededhere }
```

tests/unit/abilities/report-test.js
```diff
@@ -1,36 +1,46 @@
...
 module("Unit | Ability | report", function (hooks) {
   setupTest(hooks);
+  setupSession(hooks);
 
   test("can edit when user is superuser", function (assert) {
     const ability = this.owner.lookup("ability:report");
-    ability.set("user", EmberObject.create({ isSuperuser: true }));
+    const session = this.owner.lookup("service:session");
+    session.data = { user: EmberObject.create({ isSuperuser: true }) };
+    // ability.set("user", EmberObject.create({ isSuperuser: true }));
```
derrabauke, line 14: Delete comments

derrabauke (approved):
*bueno* :+1:

Issue #911, opened 2023-04-20
Changing project/task in activity bar


trowik:
There's a bug when starting an activity after selecting a most recent task and changing it.
This results in fetching wrong tasks (tasks for most recent project and not the selected one).
Happens only when selecting most recent task and only for activities, time sheet works fine

How to replicate:
1. select a most recent task
2. change the project
4. tasks for most recent project are loaded instead of the selected one

derrabauke:
Hey @Melkor333, I just tested it and you are right. If you stop the activity from the activities table (not directly in the task-selection on top of the app), then the task-selection won't reset properly. Further selecting another *recent* task does not fill the `project` properly, but it does fill in the `customer` and `task` right. When start the selected activity from this state, it will correct the `project` as soon as you hit the start button.

So it's confusing for a second, but it won't track a wrong task or lead to data loss AFAICT.

We will try to fix that!

trowik:
Tested it yesterday and looks good to me

PR #998, opened 2023-06-27
fix: changing project in the activity bar when change the recent customer

MitanOmar:


MitanOmar:
final fix of #911

app/components/task-selection/component.js
```diff
@@ -338,7 +366,11 @@ export default class TaskSelectionComponent extends Component {
   onTaskChange(value, options = {}) {
     this._task = value;
 
-    if (!this.project && value?.get("project.id")) {
+    const projectId = value?.get("project.id");
+    if (
+      (!this.project && projectId) ||
+      (projectId && this.project?.id !== projectId)
+    ) {
```
derrabauke, line 373: @MitanOmar, I covered the reset of the `project` property in this handler, since it will be called anyways. So we can keep our code more *DRY*.

derrabauke (approved):

PR #1018, opened 2023-07-18
chore: remove ember parachute

MitanOmar:
remove ember parachute

### DIFF
--- a/app/analysis/edit/controller.js
+++ b/app/analysis/edit/controller.js
@@ -4,42 +4,21 @@ import { action } from "@ember/object";
 import { later } from "@ember/runloop";
 import { inject as service } from "@ember/service";
 import { dasherize } from "@ember/string";
+import { tracked } from "@glimmer/tracking";
 import { task } from "ember-concurrency";
 import {
   underscoreQueryParams,
-  serializeParachuteQueryParams,
+  serializeQueryParams,
   filterQueryParams,
+  getDefaultQueryParamValue,
+  allQueryParams,
 } from "timed/utils/query-params";
+import {
+  serializeMoment,
+  deserializeMoment,
+} from "timed/utils/serialize-moment";
 import { cleanParams, toQueryString } from "timed/utils/url";
 import IntersectionValidations from "timed/validations/intersection";
-
-import { AnalysisQueryParams } from "../index/controller";
-
-/* eslint-disable ember/avoid-leaking-state-in-ember-objects */
-export const AnalysisEditQueryParams = AnalysisQueryParams.extend({
-  id: {
-    defaultValue: [],
-    replace: true,
-    refresh: true,
-    serialize(arr) {
-      return (arr && arr.join(",")) || null;
-    },
-    deserialize(str) {
-      return (str && str.split(",")) || [];
-    },
-  },
-});
-/* eslint-enable ember/avoid-leaking-state-in-ember-objects */
-
-const prepareParams = (params) =>
-  cleanParams(
-    underscoreQueryParams(
-      serializeParachuteQueryParams(
-        filterQueryParams(params, "editable"),
-        AnalysisEditQueryParams
-      )
-    )
-  );
 
 const filterUnchanged = (attributes, changes) => {
   return Object.keys(attributes).reduce((obj, attr) => {
@@ -53,9 +32,28 @@ const TOOLTIP_CANNOT_VERIFY =
   "Please select yourself as 'reviewer' to verify reports.";
 const TOOLTIP_NEEDS_REVIEW = "Please review selected reports before verifying.";
 
-export default class AnalysisEditController extends Controller.extend(
-  AnalysisEditQueryParams.Mixin
-) {
+export default class AnalysisEditController extends Controller {
+  queryParams = [
+    "customer",
+    "costCenter",
+    "project",
+    "task",
+    "user",
+    "reviewer",
+    "billingType",
+    "costCenter",
+    "fromDate",
+    "toDate",
+    "review",
+    "notBillable",
+    "verified",
+    "billed",
+    "ordering",
+    "editable",
+    "rejected",
+    "id",
+  ];
+
   IntersectionValidations = IntersectionValidations;
 
   @service notify;
@@ -64,6 +62,24 @@ export default class AnalysisEditController extends Controller.extend(
   @service session;
   @service store;
   @service unverifiedReports;
+
+  @tracked id;
+  @tracked user;
+  @tracked reviewer;
+  @tracked customer;
+  @tracked project;
+  @tracked task;
+  @tracked billingType;
+  @tracked costCenter;
+  @tracked fromDate;
+  @tracked toDate;
+  @tracked review;
+  @tracked notBillable;
+  @tracked verified;
+  @tracked billed;
+  @tracked editable;
+  @tracked rejected;
+  @tracked ordering = "-date";
 
   get analysisIndexController() {
     return getOwner(this).lookup("controller:analysis.index");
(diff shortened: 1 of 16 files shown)


app/analysis/edit/controller.js
```diff
@@ -13,49 +15,51 @@ import {
...
+
+const serializeMoment = (momentObject) => {
+  if (momentObject) {
+    momentObject = moment(momentObject);
+  }
+  return (momentObject && momentObject.format(DATE_FORMAT)) || null;
+};
+
+const deserializeMoment = (momentString) =>
+  (momentString && moment(momentString, DATE_FORMAT)) || null;
```
derrabauke, line 35: Could we externalize this into a util/helper function?

app/analysis/index/controller.js
```diff
@@ -143,14 +70,97 @@ export default class AnalysisController extends Controller.extend(
...
+  @action
+  updateVerified(value) {
+    this.verified = value;
+    this._reset();
+  }
+
+  @action
+  updateBilled(value) {
+    this.billed = value;
+    this._reset();
```
derrabauke, line 144: This does look cumbersome and error prone. Let's keep it `DRY`. I think what we need here is some sort of `tracked` function which is triggered whenever one of those query params change.

app/analysis/index/controller.js
```diff
@@ -462,4 +460,69 @@ export default class AnalysisController extends Controller.extend(
...
+      const key = Object.keys(param)[0];
+      queries[key] = this[key];
+    }
+    return queries;
+  }
+
+  get queryParamsState() {
+    const states = {};
+    for (const param of this.queryParams) {
+      // the type of param is string of object that have one key
```
derrabauke, line 481: Could you clarify this comment? What is that *special case* you are handling there with the `Object.keys(param)[0]`?

app/analysis/index/controller.js
```diff
@@ -462,4 +460,69 @@ export default class AnalysisController extends Controller.extend(
...
+        states[key].deserialize = (val) => {
+          return val.replace(",id", "");
+        };
+      }
+    }
+    return states;
+  }
+
+  getDefaultValue(paramKey) {
+    for (const param of this.queryParams) {
```
derrabauke, line 509: This method seems a little useless. It's returning `undefined` and `Object.values(param)[0];` for what reason?

derrabauke: May [trackedReset](https://github.com/tracked-tools/tracked-toolbox#trackedreset) could be helpful here.

pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.1'
+lockfileVersion: '6.0'
```
derrabauke, line 1: Update your `pnpm` and `install` again. Downgrading the `lock` file should not happen.

tests/acceptance/analysis-edit-test.js
```diff
@@ -94,7 +94,7 @@ module("Acceptance | analysis edit", function (hooks) {
   });
 
   test("cannot verify unreviewed reports", async function (assert) {
-    await visit("/analysis/edit?id=1,2,3");
+    await visit('/analysis/edit?id=%5B"1"%2C"2"%2C"3"%5D');
```
derrabauke, line 97: Boy, this looks ugly :sweat_smile: Why did the query params change here?

tests/unit/utils/query-params-test.js
```diff
@@ -9,10 +9,8 @@ module("Unit | Utility | query params", function () {
   test("can serialize parachute query params", function (assert) {
```
derrabauke, line 9: We want to get rid of *parachute* in **all** the places. ;)

app/utils/query-params.js
```diff
@@ -32,7 +32,7 @@ export const underscoreQueryParams = (params) => {
  */
 export const serializeParachuteQueryParams = (params, queryParamsObject) => {
```
derrabauke, line 33: We want to get rid of parachute in all the places. ;)

derrabauke (changes requested):

app/analysis/index/controller.js
MitanOmar: is string `OR` object that have one key

derrabauke: I could imagine that ~of~ **or** typo. But as a reader of your code I want to know why we have to do this *special case* handling here. What query param requires this and why?

MitanOmar: there is something wrong that i understand from the documentation, i thought that default query parmas value should be there

app/analysis/index/controller.js
derrabauke: or at least write some generic function like 
```js
@action
updateParam(key, value) {
  this[key] =  value;
  this._reset();
}
```

app/statistics/controller.js
```diff
@@ -266,15 +190,63 @@ export default class StatisticsController extends Controller.extend(
...
     );
   }
+
+  get queryParamsState() {
+    const states = {};
+    for (const param of this.queryParams) {
+      const defaultValue = getDefaultQueryParamValue(param);
+      const currentValue = this[param];
+      states[param] = {
+        as: param,
```
derrabauke, line 232: the `as` is an parachute artifact

app/statistics/controller.js
```diff
@@ -266,15 +190,63 @@ export default class StatisticsController extends Controller.extend(
...
+        value: currentValue,
+        changed: currentValue !== defaultValue,
+      };
+      if (["fromDate", "toDate"].includes(param)) {
+        states[param].serialize = serializeMoment;
+        states[param].deserialize = deserializeMoment;
+      }
+    }
+    return states;
+  }
```
derrabauke, line 244: Export the contents of this getter to the `utils/query-param.js` file as well and maybe give it a `dates` parameter so you can dynamically adapt the *qp*s which need date formatting.

app/statistics/controller.js
```diff
@@ -266,15 +190,63 @@ export default class StatisticsController extends Controller.extend(
...
+
+  resetQueryParams(...args) {
+    const params = [...args[0]];
```
derrabauke, line 247: Same as above
> Why spreading the args here and afterwards collecting it again? Just use a simple parameter in the first place.

app/analysis/edit/controller.js
```diff
@@ -226,4 +236,54 @@ export default class AnalysisEditController extends Controller.extend(
...
+    return states;
+  }
+
+  resetQueryParams(...args) {
+    if (!args[0]) {
+      return;
+    }
+    const params = [...args[0]];
+    for (const param of params) {
+      this[param] = getDefaultQueryParamValue(param);
```
derrabauke, line 275: Streamline with the other `resetQueryParams` methods. This is something you could extract into a general controller `QueryParamsController` since it's all the same.

app/analysis/edit/controller.js
```diff
@@ -226,4 +236,54 @@ export default class AnalysisEditController extends Controller.extend(
...
+        states[param].serialize = serializeMoment;
+        states[param].deserialize = deserializeMoment;
+      }
+
+      if (param === "id") {
+        states[param].serialize = (arr) => {
+          return (arr && Array.isArray(arr) && arr.join(",")) || null;
+        };
+        states[param].deserialize = (str) => {
+          return (str && str.split(",")) || [];
```
derrabauke, line 262: Extract.

derrabauke (changes requested):
We are slowly getting there! :smile_cat:

app/statistics/controller.js
MitanOmar: i was making sure to make exact replacement of ember parachute
i remember that u ask me before about why it's returning `undefined`, that's because ember parachute was returning that too

app/analysis/index/controller.js
```diff
@@ -453,13 +394,20 @@ export default class AnalysisController extends Controller.extend(
...
   reset() {
-    this.resetQueryParams(
-      Object.keys(this.allQueryParams).filter((k) => k !== "ordering")
+    resetQueryParams(
+      this,
+      Object.keys(allQueryParams(this)).filter((k) => k !== "ordering")
     );
   }
+
+  // i am puting it here in a getter, becuase it's used in the template
```
derrabauke, line 409: Comment not needed. I think this is obvious and it has a typo in it. :) 

After that we can merge it! :+1:

derrabauke (approved):
Good job :+1:

PR #1067, opened 2023-09-27
chore: ember 4.4 upgrade

derrabauke:
This is the final upgrade to `ember v4.4`.


derrabauke:
We should replace the current patched version of `ember-nouislider` with a fork in the https://github.com/adfinis-forks/ namespace. But currently I have no access to that, so I will postpone this.

MitanOmar (approved):

Issue #1055, opened 2023-09-04
Filter wrong when navigating from rejected to non verified reports


madpat1998:
When first navigating to "rejected reports" via the dedicated button on the top right and then to "unverified reports" also using the buton on the top right, no report is displayed as the filter "rejected" is still set to "rejected".

PR #1084, opened 2023-10-30
fix(report-review-warning): `URL` in the component

MitanOmar:
fix of issue #1055

### DIFF
--- a/app/components/report-review-warning/template.hbs
+++ b/app/components/report-review-warning/template.hbs
@@ -5,10 +5,11 @@
         @route="analysis.index"
         @query={{hash
           fromDate=null
-          toDate=this.unverifiedReports.reportsToDate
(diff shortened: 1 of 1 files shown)


app/components/report-review-warning/template.hbs
```diff
@@ -5,10 +5,11 @@
         @route="analysis.index"
         @query={{hash
           fromDate=null
-          toDate=this.unverifiedReports.reportsToDate
+          toDate=this.formatedToDate
```
derrabauke, line 8:
```suggestion
          toDate={{moment-format this.unverifiedReports.reportsToDate "YYYY-MM-DD"}}
```

You could try, if this works the same?

derrabauke (changes requested):

Issue #1104, opened 2023-12-12
 Multiple clicks on save creates multiple reports


trowik:
## Expected behaviour

Multiple clicks on save only creates one report
## Current behaviour

Multiple clicks on save creates multiple reports
## Steps to reproduce

1. Select task for a new report
2. Click the save button multiple times quickly
3. Refresh page

PR #1157, opened 2023-12-21
fix(report-row): components from duplication when clicking submit

MitanOmar:
fix of #1104

### DIFF
--- a/app/components/report-row/component.js
+++ b/app/components/report-row/component.js
@@ -29,7 +29,9 @@ export default class ReportRowComponent extends Component {
    */
   @dropTask
   *save(changeset) {
-    yield this.args.onSave(changeset);
+    if (this.args.onSave) {
+      yield this.args.onSave(changeset);
+    }
   }
   /**
    * Delete the row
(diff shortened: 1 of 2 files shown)


app/components/report-row/component.js
```diff
@@ -27,8 +29,10 @@ export default class ReportRowComponent extends Component {
    * @public
    */
   @action
-  save(changeset) {
-    this.args.onSave(changeset);
+  async save(changeset) {
```
derrabauke, line 32: use `dropTask` for that and you won't need any tracking variable. Further add `this.save.isRunning` to the `disabled` state check in the template.

app/components/report-row/component.js
```diff
@@ -26,9 +27,9 @@ export default class ReportRowComponent extends Component {
    * @method save
    * @public
    */
-  @action
-  save(changeset) {
-    this.args.onSave(changeset);
+  @dropTask
+  *save(changeset) {
+    yield this.args.onSave(changeset);
```
derrabauke, line 32:
```suggestion
    if(this.args.onSave) {
      yield this.args.onSave(changeset);
    }
```

Since we switched to a task we now use the `perform` helper, but this won't check if the passed argument is a function or not. (...well technically it does, but it will only yield an assertion error at in the dev env and an error on prod) Earlier we had the `optional` helper for it. So now you have to check by your own if this is a valid function or at least if it's existing.

derrabauke (approved):

PR #1159, opened 2024-01-04
fix(worktime-balance-chart): the float number was not rendered in the chart if it's long float

MitanOmar:
fix of #1158


app/components/worktime-balance-chart/component.js
```diff
@@ -17,7 +17,7 @@ export default class WorktimeBalanceChart extends Component {
       datasets: [
         {
           data: this.args.worktimeBalances.map(({ balance }) =>
-            balance.asHours()
+            balance.asHours().toFixed(2)
```
derrabauke, line 20:
```suggestion
            Number.parseFloat(balance.asHours().toFixed(2))
```
This will remove all the insignificant trailing zeros. And then you won't have to change the test assertions.

MitanOmar: done

derrabauke (approved):
:+1:

PR #1163, opened 2024-01-10
fix: verify rejected reports

### LINKED ISSUE #1091: bug: can verify rejected reports
you can verify rejected reports, shouldn't be like that

MitanOmar:
fix #1091


derrabauke (approved):
In harmony with the [backend](https://github.com/adfinis/timed-backend/blob/5d8cef45d9687892b122631d639d6d5c239ff61d/timed/tracking/models.py#L103) :notes:

PR #1174, opened 2024-02-13
feat: enforce spelling check for comment field

### LINKED ISSUE #1173: Enforce spelling check for "Comment" field
The text in the "Comment" field is important because it contains text that customers may see. I often make silly typos that could be eliminated by using the browser's spell checker. However, spell checking is not enabled by default for this field.

Could we add spellcheck="true" to this field?

MitanOmar:
fix #1173


derrabauke (approved):
:+1:

PR #1186, opened 2024-03-13
feat: highlight rows in analysis with the review flag

StephanH90:
I routinely select a bunch of rows during review, click "edit selected reports" and then realise that i cant verify them because one of them had the review flag set and now i need to reselect them all. 

This is an attempt to alleviate this pain:

image

A nicer solution would be to track the selection of the reports in the url, but that could potentially break a lot of business logic.

### DIFF
--- a/app/analysis/index/template.hbs
+++ b/app/analysis/index/template.hbs
@@ -350,9 +350,12 @@
                   {{#each reports as |report|}}
                     {{! template-lint-disable}}
                     <tr
-                      class="{{if
-                          (includes report.id this.selectedReportIds)
-                          'selected'
+                      class="{{concat
+                          (if
+                            (includes report.id this.selectedReportIds)
+                            'selected'
+                          )
+                          (if report.review ' needs-review')
                         }}
                         {{if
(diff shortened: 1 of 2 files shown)


anehx:
> A nicer solution would be to track the selection of the reports in the url, but that could potentially break a lot of business logic.

Actually it is tracked but if the bulk edit is canceled we don't pass it on.. That would be a nice feature but I also like this solution!

However, I'm not sure if we need to use a separate color if the "needs review"-report is selected..

derrabauke:
Actually the selection was remaining until we have refactored the whole app/analysis controller and by doing that the feature got lost. 

https://github.com/adfinis/timed-frontend/blame/d0cb64e0b9f8bd37d5b92144127dc5b1b5dbac80/app/analysis/edit/controller.js#L166

Will try to restore that functionality!

derrabauke:
> Will try to restore that functionality!

Done here https://github.com/adfinis/timed-frontend/pull/1187

StephanH90:
@derrabauke the real MVP.

PR #1199, opened 2024-04-16
fix(report-abilities): get function is not exists

MitanOmar:


app/components/sy-topnav/template.hbs
```diff
@@ -61,7 +61,7 @@
       <li class="nav-top-list-item">
         <LinkTo @route="users.edit" @model={{this.session.data.user.id}}>
           <FaIcon @icon="user" />
-          {{this.session.data.user.fullName}}
```
StephanH90, line 64: before we had fullName on the user and now thats gone? How come?

app/components/sy-topnav/component.js
```diff
@@ -12,4 +12,23 @@ export default class SyTopnav extends Component {
...
+   */
+  get fullName() {
+    const fullName = this.session.data.user?.fullName;
+    if (fullName) return fullName;
+
+    if (!this.session.data.user?.attributes) return "";
+
+    const userAttributes = this.session.data.user.attributes;
+    return `${userAttributes["first-name"]} ${userAttributes["last-name"]}`;
+  }
```
StephanH90, line 33: I would first check if I really need this function. Otherwise you might be able to shorten it to:
```suggestion
  get fullName() {
    return this.session?.data?.user?.fullName ?? "${this.session?.data?.user?["firstName"]} ${this.session?.data?.user?["last-name"]}"
  }
```

StephanH90 (changes requested):

app/components/sy-topnav/component.js
```diff
@@ -12,4 +12,23 @@ export default class SyTopnav extends Component {
...
+
+  /**
+   * TODO:
+   * i don't know if it's an issue in `ember-simple-auth` or `ember-simple-auth-oidc`
+   * or it's maybe wrong configration from Timed frontend side,
+   * BUT after refreshing the token, the session service is not encapsulating the reslts from keycloak
+   * inside ember data model, so i add this small.
+   *
+   * if the issue is fixed, change everything in app/abilities/report.js that belongs to `this.user?.id`
+   */
```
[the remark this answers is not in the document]
MitanOmar: because in `app/abilities/report.js`, i remove `.get` so if the issue is fixed, it will be better to return it again to use `.get` to access the attribute

app/components/sy-topnav/template.hbs
MitanOmar: because there is an issue with `ember-simple-auth` or `ember-simple-auth-oidc`,
because in the first time login, the `session.data` is return an instance of ember data model, but after refreshing the token, it's return just an object of the user information, so i can not use the helper that is defined in `User` data model

app/components/sy-topnav/component.js
Yelinz: thats a lot of question marks :rofl:

MitanOmar: it was able for me to do that, but i wanted it to be more clear and readable :)

app/components/sy-topnav/component.js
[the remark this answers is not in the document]
Yelinz: Shouldnt get anways get removed? In newer ember-data versions get is not necessary to retrieve attributes which are not relationships or is user not from ember-data.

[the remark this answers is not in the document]
MitanOmar: then, i will remove the comment that belong to replacing report abilities part

